PR #12, opened 2018-07-06
Extra services: download formats, refs/cits, DBLP, ancillary files (ARXIVNG-597, ARXIVNG-943, ARXIVNG-956)

mhl10:
This PR bundles together three extra-services features and corresponding tests:
* Reference and citation section (ARXIVNG-957)
* DBLP Bibliography linking (ARXIVNG-956)
* ancillary file display (ARXIVNG-943)

None of these are particularly interesting features. The DBLP feature leverages the DB service to perform lookups but also offers a fallback option if the DB service isn't available. The fallback option works, but the functionality around DB service availability hasn't been implemented yet (i.e. right now we assume it's always available).

Examples in test data:
* 1502.00001 demonstrates the reference and citation section with INSPIRE links (all papers are intended to show the NASA ADS link)
* 0704.0361 demonstrates the DBLP linking in the test data. 
* For the ancillary file display, I used the source tarball from 0905.2326 to test, but I have *not* committed those to the repository. It would probably be OK to include in the test data, since there's no personal information included, but the tarball is nearly 3MB. Should I just included some stub data instead?

Please note that Travis is still returning a failure with this PR; those problems will be addressed separately.


browse/services/document/author_affil.py
```diff
@@ -270,18 +283,21 @@ def _parse_author_affil_back_propagate(author_list: List[List[str]],
     return author_list
 
 
-def split_authors(authors: str)->List:
-    """Split author string into authors entity lists.
+def split_authors(authors: str) -> List:
```
bbarker: Not a comment on the changes in this PR. But, this is a fairly complicated bit of parsing going on that might be simplified by 1) writing a very simple recursive descent parser, 2) not parsing in the future, if we no longer work with author lists. Hopefully (2), as I'm not sure recursive descent is an easy match for python, without looking into it more.

browse/services/document/metadata.py
```diff
@@ -325,6 +329,97 @@ def get_previous_id(self, identifier: Identifier):
...
+        # Separate check for ScienceWISE annotated PDF
+        if add_sciencewise:
+            if formats and formats[-1] == 'other':
+                formats.insert(-1, 'sciencewise_pdf')
+            else:
+                formats.append('sciencewise_pdf')
+
+        return formats
+
+    def get_ancillary_files(self, docmeta: DocMetadata) \
```
bbarker: Is there logic here for why an `Option[List[_]]` is needed? Generally, it is ok to return an empty List in place of None, but there could be exceptions if we should really treat an empty list differently from `None` in the calling code.

This comment applies to several other places in the code as well.

browse/routes/ui.py
```diff
@@ -37,7 +37,11 @@ def bare_abs():
 @blueprint.route('/abs/<path:arxiv_id>', methods=['GET', 'POST'])
 def abstract(arxiv_id: str) -> Union[str, Response]:
     """Abstract (abs) page view."""
-    response, code, headers = abs_page.get_abs_page(arxiv_id, request.args)
+    download_format_pref = request.cookies.get('xxx-ps-defaults')
+    print(f'download format pref: {download_format_pref}')
```
bdc34: This might be a debugging statement left in the PR.

browse/services/document/author_affil.py
```diff
@@ -102,7 +106,7 @@ def _parse_author_affil_split(author_line: str)->Dict:
                  r'^(.*)\s+(' + PREFIX_MATCH + r')\s(' +
                  PREFIX_MATCH + r')\s(\S+)$'),
                 ('name-prefix-name',
-                 r'^(.*)\s+(' + PREFIX_MATCH + ')\s(\S+)$'),
+                 r'^(.*)\s+(' + PREFIX_MATCH + r')\s(\S+)$'),
```
bdc34: Was the lack of the the r here causing errors or is this a style change?

mhl10: I think all this does is silence the linter, which just complains about invalid escape sequences in the string.

browse/services/document/metadata.py
mhl10: There was really no need to return `None` in all of the cases I had used `Optional[List[_]]` so I updated the functions to return `List[_]`.

PR #13, opened 2018-07-09
Submitter view email link generation (ARXIVNG-596)

mhl10:
This probably should have been included with https://github.com/cul-it/arxiv-browse/pull/12 since the changes here are minor, but it's worth having a look at the changes to see what I'm doing about the secret with respect to the link generation here.


browse/services/document/metadata.py
```diff
@@ -607,8 +623,10 @@ def get_session(app: object = None) -> AbsMetaSession:
     config = get_application_config(app)
     orignal_versions_path = config.get('DOCUMENT_ORIGNAL_VERSIONS_PATH', None)
     latest_versions_path = config.get('DOCUMENT_LATEST_VERSIONS_PATH', None)
+    show_email_secret = config.get('SHOW_EMAIL_SECRET', '')
```
bbarkerro: Is this normally determined by a per-user entry in the DB?

mhl10: It's an application-level variable.

browse/services/document/metadata.py
```diff
@@ -568,12 +577,19 @@ def _parse_metadata_fields(fields: Dict, key_value_block: str) -> None:
         return fields
 
 
+
+@wraps(AbsMetaSession.get_show_email_hash)
+def get_show_email_hash(identifier: Identifier) -> str:
```
bdc34: Could the email hash be only at the UI level? It doesn't really have anything to do with the metadata. The whole email hash might be just a UI feature.

browse/services/search/search_authors.py
```diff
@@ -8,37 +7,44 @@
 
 
 def is_affiliation(item: str)-> bool:
+    """Return true if a string contains an affiliation."""
     return item.startswith('(')
 
 
 def is_short(item: str)-> bool:
+    """Return true if the length of string is less than 4 characters long."""
```
bdc34: These comments seem redundant. Is this just to achieve a pylint score?

mhl10: pydocstyle checks complain (fail?) otherwise.

bdc34: Ah.

mhl10:
@bdc34 I just pushed some changes that use the UI (filter) based approach. Does this seems like a reasonable interim solution for building the /show-email URL?

browse/services/document/metadata.py
mhl10: I just pushed some changes that use the UI (filter) based approach. Does this seems like a reasonable interim solution for building the /show-email URL?

bdc34: Looks good, can you remove the remaining email hash stuff in metadata.py? It looks like it doesn't need SHOW_EMAIL_SECRET any more.

PR #15, opened 2018-07-11
Social bookmarks and clickthrough

bdc34:


tests/test_click_through.py
```diff
@@ -0,0 +1,15 @@
...
+
+from hypothesis import given
+from hypothesis.strategies import text
+
+from browse.domain.clickthrough import create_hash, is_hash_valid
+
+
+class TestClickthrough(unittest.TestCase):
+
+    @given(text(), text())
```
mhl10: 👍 It's nice to see property-based testing; hypothesis seems like a solid library for doing this!

browse/domain/clickthrough.py
```diff
@@ -0,0 +1,28 @@
+"""Functions to create hashes for Click through links.
```
mhl10: I think this should be moved out of domain and into a util directory.

bdc34: I've moved this.

PR #17, opened 2018-07-18
Adds some tests. Improves logging.

bdc34:
This PR
-  Adds some tests to improve coverage
- Removes a test file that was totally commented out.
- Simplifies the use of the get institution functions. 
- Changes abs controller to not fail if the creation of non-critical data throws exceptions.


browse/controllers/abs_page/__init__.py
```diff
@@ -77,24 +78,31 @@ def get_abs_page(arxiv_id: str,
...
+            # Ancillary files
+            response_data['ancillary_files'] = \
+                metadata.get_ancillary_files(abs_meta)
+
+            # Browse context
+            _check_context(arxiv_identifier,
+                           request_params,
+                           response_data)
+        except Exception as e:
+            app.logger.error("Error getting non-critical abs page data", exc_info=app.debug)
```
mhl10: Could we use logging from arxiv-base for consistency with the other repos? e.g.
```
from arxiv.base import logging
logger = logging.getLogger(__name__)
...
logger.error(f'Error getting non-critical abs page data: {e}')
```
See also https://github.com/cul-it/arxiv-base/blob/master/arxiv/base/logging.py

browse/controllers/abs_page/__init__.py
```diff
@@ -120,11 +128,9 @@ def get_abs_page(arxiv_id: str,
     except IdentifierException:
         raise AbsNotFound(data={'arxiv_id': arxiv_id})
     except (AbsException, Exception) as e:
-        print(f'Problem: {e}')
```
mhl10: It would probably be good to log this too.

browse/routes/ui.py
```diff
@@ -16,8 +17,7 @@
 def before_request() -> None:
     """Get instituional affiliation from session."""
     if 'institution' not in session:
-        institution = get_institution_from_request()
-        session['institution'] = institution
+        session['institution'] = get_institution(request.remote_addr)
```
mhl10: This change seems OK--more straightforward. I wonder if this should be handling exceptions though?

browse/services/database/__init__.py
```diff
@@ -45,8 +44,6 @@ def get_institution(ip: str) -> Optional[str]:
         return institution_name  # type: ignore
     except NoResultFound:
         return None
-    except SQLAlchemyError as e:
```
mhl10: Curious why this was removed from the DB service functions?

browse/controllers/abs_page/__init__.py
```diff
@@ -7,6 +7,7 @@
 
 from typing import Tuple, Dict, Any, Optional
 from flask import url_for
+from flask import current_app as app
```
bbarker: Good to see this is how we are accessing logging; I wasn't sure before.

bbarker: Nevermind, I see @mhl10's comment below on `arxiv.base import logging`

tests/test_browse.py
```diff
@@ -59,6 +63,13 @@ def test_abs_with_truncated_author_list(self):
...
+    def test_all_abs_as_web_pages(self):
+        num_files_tested = 0
+        for dir_name, subdir_list, file_list in os.walk(ABS_FILES):
+            for fname in file_list:
+                fname_path = os.path.join(dir_name, fname)
+                if os.stat(fname_path).st_size == 0 or not fname_path.endswith('.abs'):
+                    continue
+                m = AbsMetaSession.parse_abs_file(filename=fname_path)
+                rv = self.app.get(f'/abs/{m.arxiv_id}')
+                self.assertEqual(rv.status_code, 200)
```
bbarker: Nice test! Other than checking for status 200, would it be worth checking for particular bits of common content?

tests/test_database_service.py
```diff
@@ -204,6 +208,35 @@ def test_get_dblp_authors(self) -> None:
...
+        self.assertEqual(
+            TestBrowseDatabaseService.database_service.count_all_trackback_pings(), 0)
+        self.assertEqual(
+            TestBrowseDatabaseService.database_service.has_sciencewise_ping('0704.0361'), False)
+        self.assertEqual(
+            TestBrowseDatabaseService.database_service.get_dblp_listing_path('0704.0361'), None)
+        self.assertEqual(
+            TestBrowseDatabaseService.database_service.get_dblp_authors('0704.0361'), [])
+        mock_query.side_effect = SQLAlchemyError
+        self.assertRaises(SQLAlchemyError,
```
mhl10: 👍 Nice, I didn't see this before! Still curious about dropping the SQLAlchemyError handling though :)

tests/test_formats.py
```diff
@@ -33,3 +33,28 @@ def test_formats_from_source_type(self):
         self.assertListEqual(
             formats_from_source_type('', format_pref='fname=CM'),
             ['ps(CM)', 'other'])
+        self.assertListEqual(formats_from_source_type('P'),
```
mhl10: Thanks for adding these!

bdc34: Your welcome.

browse/services/database/__init__.py
bdc34: Transforming the SQLAlchemyError to an IOError seems arbitrary and doesn't add anything. They both should get handled by the arxiv-base 500 handler.

tests/test_database_service.py
bdc34: Each path through exception handling code requires additional testing. So letting as many types of exceptions go to arxiv-base 500 as possible makes it so we can test fewer types of exceptions. Almost all will be handled uniformly by arxiv-base's 500 and we know we have that tested.

browse/controllers/__init__.py
```diff
@@ -4,22 +4,3 @@
...
-        institution_str = get_institution(request.remote_addr)
-
-    except IOError:
-        # TODO: log this
-        # return {
-        #     'explanation': 'Could not access the database.'
-        # }, status.HTTP_500_INTERNAL_SERVER_ERROR, {}
-        return None
-
-    return institution_str
```
bdc34: @mhl10 I moved this to the abs controller but I think we might want to move this to base since other pages will need the institution header?

browse/controllers/abs_page/__init__.py
bdc34: I think that base is going to set the logger on app during the call to Base(app) and that this is all we need to do. I might be wrong, I'll check.

bdc34: I've changed this to use the arxiv.base logging pattern above.

browse/controllers/abs_page/__init__.py
bdc34: This gets changed in another branch to go through to the arxiv-base 500 handler.

browse/routes/ui.py
bdc34: Goes to arxiv-base 500. It's less code and doesn't need a test. 

get_institution() might have to go arxiv-base if other pages need that institutional banner too.

browse/controllers/abs_page/__init__.py
```diff
@@ -6,19 +6,21 @@
 """
 
 from typing import Tuple, Dict, Any, Optional
-from flask import url_for
 from urllib.parse import urljoin
+
+from flask import url_for
+from flask import current_app as app
```
bbarker: probably don't need the 'app' import anymore?

PR #19, opened 2018-07-24
Type improvements and related logic fixes

bbarker:
The lack of local feedback during development has resulted in an accumulation of type errors. I've fixed the existing type errors in the browse module proper, and reinstated a version of the old test script.

As previously discussed, we decided to not put an emphasis on types in test code, but we have nearly 150 errors there - I've capped it arbitrarily at 150 in the test script. However, any errors are still spit out.

We can discuss how to best get feeback by perhaps changing the `MyPyTest` class to suit our needs.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-arxiv-base = "*"
+arxiv-base = ">=0.9.1"
 dataclasses = "*"
 inflect = "*"
 python-dateutil = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-arxiv-base = "==0.9.1"
+arxiv-base = "*"
 dataclasses = "*"
 inflect = "*"
 python-dateutil = "*"
@@ -14,7 +14,6 @@ Flask-SQLAlchemy = "*"
 "jinja2" = "==2.10"
 MarkupSafe = "*"
 SQLAlchemy = "*"
-werkzeug = "==0.13"
 pytz = "*"
 pylint = "*"
 coverage = "*"
@@ -25,8 +24,8 @@ mysqlclient = "*"
 
 [dev-packages]
 pylama = "*"
-mypy = "==0.560"
-"nose2" = "==0.7.3"
+mypy = "*"
+"nose2" = "*"
 sqlacodegen = "*"
 Flask-Testing = "*"
 pycodestyle = "*"
--- a/browse/domain/identifier.py
+++ b/browse/domain/identifier.py
@@ -101,12 +101,11 @@ class Identifier:
                     'invalid arXiv identifier {}'.format(self.ids)
                 )
         self.has_version: bool = False
+        self.idv: str = self.id
         if id_match.group('version'):
             self.version = int(id_match.group('version'))
             self.idv = f'{self.id}v{self.version}'
             self.has_version = True
-        else:
-            self.idv = self.id
         self.squashed = self.id.replace('/', '')
         self.squashedv = self.idv.replace('/', '')
         self.yymm: str = id_match.group('yymm')
--- a/browse/domain/identifier.py
+++ b/browse/domain/identifier.py
@@ -1,7 +1,7 @@
 """Base domain classes for browse service."""
 import json
 import re
-from typing import Match
+from typing import Dict, Match, Optional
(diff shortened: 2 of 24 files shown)


Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-arxiv-base = "==0.9.1"
+arxiv-base = "*"
```
bdc34: 0.9.1 or greater is required.

browse/domain/identifier.py
```diff
@@ -40,7 +40,10 @@ class IdentifierIsArchiveException(IdentifierException):
     pass
 
 
-class Identifier():
+I = TypeVar('I', bound='Identifier')
```
bdc34: What is this for?

browse/services/util/metatags.py
```diff
@@ -34,7 +34,8 @@ def meta_tag_metadata(metadata: DocMetadata)->List:
         meta_tags.append(_mtag('citation_title', metadata.title))
 
     if metadata.authors:
-        hundo = parse_author_affil_utf(metadata.authors)[:100]
+
```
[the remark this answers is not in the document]
bbarkerro: @bdc34 I'm a bit stumped by this one; `parse_author_affil_utf` takes a `str` but `metadata.authors` is an `AuthorList` which has a single field, `raw` that is a `str`. For a while I was instead trying to pass in `metadata.authors.raw` here but that was causing numerous *test failures*, but not type errors (iirc). So, this was a solution that seemed to work for both mypy and dynamic reality. I would like to understand why it is failing if we convert it to `metadata.authors.raw`, though.

browse/services/document/author_affil.py
```diff
@@ -78,13 +78,13 @@ def _parse_author_affil_split(author_line: str) -> Dict:
...
 
-    names = split_authors(author_line)
+    names: List[str] = split_authors(author_line)
     if not names:
         return {'author_list': [], 'back_prop': 0}
 
     names = _remove_double_commas(names)
     # get rid of commas at back
-    names = reversed(list(dropwhile(lambda x: x == ',', reversed(names))))
+    names = list(reversed(list(dropwhile(lambda x: x == ',', reversed(names)))))
```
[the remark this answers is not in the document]
bdc34: We could also #ignore it.
I'd like to convert it to list as few times as possible.

bbarkerro:
There seems to be an issue with the travis system finding a new version of the hypothesis module, can revert if necessary.

browse/services/util/metatags.py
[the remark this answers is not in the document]
bbarker: So just for sake of completeness, I'll reiterate here that we fixed the issue - metadata.authors was being replaced by a string in a part of the system that was being #type ignored (and that has also changed somewhat).

browse/services/document/metadata.py
```diff
@@ -460,38 +478,62 @@ def parse_abs_file(filename: str) -> DocMetadata:
...
-            Category(id=x) for x in categories[1:] if len(categories) > 1
-        ]
-        if 'license' in fields:
-            fields['license'] = License(recorded_uri=fields['license'])
 
-        return DocMetadata(**fields)
+        doc_license: License = \
+            License() if 'license' not in fields else License(recorded_uri=fields['license'])  # type: ignore
+
+        return DocMetadata(  # type: ignore
```
mhl10: While this is clearer than `return DocMetadata(**fields)`, it's problematic. All the fields that have been commented out with question marks (`# journal_ref=???`) are no longer included.

mhl10: @bbarker this PR was merged before I had a chance to review it, but I wanted to draw your attention to this comment.

bbarker: Sorry about that - I'm out today, but will add this back in asap, possibly today. I'll attempt 1) add back all the fields without them being optional, 2) make them optional if it seems they are missing sometimes, 3) leave it as is otherwise, as I assume that means the fields don't exist.

bbarker: I guess this will also be a good chance to add some tests around these fields to check they are indeed added?

mhl10: No worries. I should have made the tests more comprehensive when I originally wrote `test_abs_parser.py`.  None of those examples actually have non-`None` values for those fields.

bbarkerro: @mhl10 working for fixes in #22

PR #22, opened 2018-07-27
Fixup parse abs file

bbarker:
This was to fix up some issues pointed out by @mhl10 in #19, where I didn't include optional members in the construction of `DocMetadata` instances. To address this, I 1) removed default factories and values for instances that *are required* so an error will be issued in case they are not, and 2) provided the values, if they exist, for optional arguments, in `parse_abs_file`.

I know @mhl10 is on vacation, just wanted to send this his way so he can see it when he is back, but probably need to merge it before then.

### DIFF
--- a/browse/domain/metadata.py
+++ b/browse/domain/metadata.py
@@ -60,7 +60,21 @@ class AuthorList:
 
 @dataclass
 class Category:
-    """Represents an arXiv category."""
+    """Represents an arXiv category.
+
+    arXiv categories are arranged in a hierarchy where there are archives
+    (asrto-ph, cs, math, etc.) that contain subject classes (astro-ph has
+    subject classes CO, GA, etc.). We now use the term category to refer
+    to any archive or archive.subject_class that one can submit to (so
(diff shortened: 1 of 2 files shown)


bdc34:
Please write a test for  `1) removed default factories and values for instances that are required so an error will be issued in case they are not,` 

Aside: I'm liking dataclasses less and less and I'm wondering if we should just go with normal classes. It's more code but less inscrutable magic.

bbarker:
@bdc34 re: dataclasses Yeah I know what you mean about dataclasses - I want to like them and mostly do, but I'm not sure to what extent the "magic" will be an issue in the long run. I think we need to be careful about adding factory defaults when it doesn't seem to make sense to do so, for instance. 

One problem right now is that tooling hasn't caught up. I have issues with both intellij and mypy - though dataclasses are supported in mypy, it is only supported in 3.7, though there is an issue to backport support to 3.6 that the devs seem to think will be easy.

bbarker:
OK, all set now I think

PR #23, opened 2018-08-03
jinja filters for URLs, arXiv IDs, DOIs and line feeds

bdc34:
Builds on Martin's work with jinja filters.
https://culibrary.atlassian.net/browse/ARXIVNG-988

### DIFF
--- a/browse/filters.py
+++ b/browse/filters.py
@@ -1,34 +1,29 @@
 """Browse jinja filters."""
 import re
 from urllib import parse
-from jinja2 import evalcontextfilter, Markup, escape
+from jinja2 import Markup, escape
 from flask import url_for
-from typing import Match
+from typing import Match, Callable
+
 
 # TODO tests for these
-# TODO escape text of <a> tags?
 
-# TODO can we use a library function for this? there must be one, Jinja escape?
-# Is it just jinja.escape?
-def _escape_special_chracters(text: str) -> str:
-    result = text
-    result = re.sub(r'&(?!amp;)', '&amp;', result)
-    result = re.sub(r'<', '&lt;', result)
-    result = re.sub(r'>', '&gt;', result)
-    return result
+def doi_urls(clickthrough_url_for: Callable[[str], str], text: str) -> str:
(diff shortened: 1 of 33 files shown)


browse/filters.py
```diff
@@ -0,0 +1,105 @@
...
+    # Two cases:
+    #  1. we get a markup object for text, ex from a previously filter stage
+    #  2. we get a raw str for text, ex meta.abstract
+    # In both of cases the value in result ends up escaped after this conditional.
+    # Then we sub DOI with HTML elements, and return the result as Markup()
+
+    if hasattr(text, '__html__'):
+        result = text
+    else:
+        result = Markup(escape(text))
```
bbarker: Haven't used this yet, but reading the docs, Markup "Marks a string as being safe for inclusion in HTML/XML output without needing to be escaped" - so why do we use the `escape`, then `Markup`? Is this just a way to be more sure that the escaping has indeed happened by specifying it manually?

tests/test_typings.py
```diff
@@ -44,7 +44,7 @@ def test_run_mypy_tests(self) -> None:
               f'{test_failures}  * * * *****\n'
               f'***** * * *   See {test_out_file_name} for error log  * * * *****'
               )
-        self.assertLessEqual(test_failures, 150, f'mypy on test {test_file}')
+        self.assertLessEqual(test_failures, 150, f'mypy run on tests')
```
bbarker: If we don't want to make this a hard test, I think we can just remove this line.

browse/filters.py
bdc34: To Jinja, a Markup object is safe to put on the page.  Most operations on a Markup object result in another Markup object and it is smart enough to not double escape. 

When we construct a Markup object, the constructor considers our argument at something we have marked up. We know it is totally safe. So if we get text into our filter that is not a Markup object (that is what that hasattr(text,'__html__') check is about), we need to escape it.

Markup( 'some text <a> stuff' ) will NOT escape the <a> 

Martin had some code that was working with autoescape but I didn't keep it.

PR #24, opened 2018-09-10
ARXINVG-1038 - add support for legacy /abs request parameters

mhl10:
This PR essentially covers the conversion of https://github.com/cul-it/arxiv-httpd/blob/master/cgi-bin/abs.pl#L63-L124. 

Some of these cases (like singleton parameters on /abs absent of any other path info) are fairly rare and might be candidates for removal in a future release. 

The `fmt=txt` case is supported by a new `Docmetadata` field called `raw_safe` which is the raw abs string minus the submitter email address.

Examples are demonstrated in test cases.

### DIFF
--- a/browse/routes/ui.py
+++ b/browse/routes/ui.py
@@ -1,4 +1,5 @@
 """Provides the user intefaces for browse."""
+import re
 from typing import Union
 
 from arxiv import status
@@ -7,6 +8,7 @@ from flask import Blueprint, render_template, request, Response, session, \
 from werkzeug.exceptions import InternalServerError, NotFound
 
 from browse.controllers import abs_page
+from browse.exceptions import AbsNotFound
 from browse.util.clickthrough import is_hash_valid
 from browse.services.database import get_institution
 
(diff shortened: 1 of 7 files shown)


browse/routes/ui.py
```diff
@@ -30,21 +31,43 @@ def apply_response_headers(response: Response) -> Response:
...
+
+    """Return abs-specific 404."""
+    raise AbsNotFound
 
 
 @blueprint.route('/abs/', methods=['GET'], defaults={'arxiv_id': ''})
-@blueprint.route('/abs/<path:arxiv_id>', methods=['GET', 'POST'])
-def abstract(arxiv_id: str) -> Union[str, Response]:
+@blueprint.route('/abs/<path:arxiv_id>', methods=['GET'])
+def abstract(arxiv_id: str) -> Response:
```
bbarker: I was wondering why it had been `Union[str, Response]` in the past... this makes more sense, I think. Also glad that using the `abstract` function to service multiple controllers worked out.

tests/test_abs_parser.py
```diff
@@ -26,13 +28,17 @@ def test_bulk_parsing(self):
                 if not fname_path.endswith('.abs'):
                     continue
                 num_files_tested += 1
-                m = AbsMetaSession.parse_abs_file(filename=fname_path)
-                self.assertIsInstance(m, DocMetadata)
-                self.assertNotEqual(m.license, None)
-                self.assertNotEqual(m.license.effective_uri, None,
+                dm = AbsMetaSession.parse_abs_file(filename=fname_path)
```
bbarker: :+1:

PR #27, opened 2018-09-18
Ensure correct canonical / non-canonical category display ARXIVNG-605

bdc34:
https://culibrary.atlassian.net/browse/ARXIVNG-605

### DIFF
--- a/browse/domain/metadata.py
+++ b/browse/domain/metadata.py
@@ -7,7 +7,7 @@ from dataclasses import dataclass, field
 from arxiv import taxonomy
 from browse.domain.identifier import Identifier
 from browse.domain.license import License
-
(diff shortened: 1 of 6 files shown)


browse/domain/metadata.py
```diff
@@ -256,3 +218,23 @@ def get_datetime_of_version(
             return None
         else:
             return versions[0].submitted_date
+        
+    def display_secondaries(self)-> List[str]:
+        """Unalias, dedup and sort secondaries for display."""
+        if not self.secondary_categories:
+            return []
+
+        def unalias(secs): # type: ignore
```
bbarker: Just out of curiosity, why the # type: ignores in this area? Maybe add a comment about why if possible.

PR #28, opened 2018-09-20
ARXIVNG-1192, ARXIVNG-1202: minor bugfixes

mhl10:
* ARXIVNG-1192 - adds a filter that wraps the tex2utf function. Filter is being applied to the title field
* ARXIVNG-1202 - adds a generic 400 template so that the /ct route can raise `BadRequest` with a message when the hash doesn't match

### DIFF
--- a/browse/config.py
+++ b/browse/config.py
@@ -224,3 +224,6 @@ DOCUMENT_CACHE_PATH = os.environ.get(
 
 # Used in linking to /show-email
 SHOW_EMAIL_SECRET = os.environ.get('SHOW_EMAIL_SECRET', 'foo')
+
+# Used in linking to /ct
+CLICKTHROUGH_SECRET = os.environ.get('CLICKTHROUGH_SECRET', 'bar')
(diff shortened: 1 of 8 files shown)


browse/config.py
```diff
@@ -224,3 +224,6 @@
 
 # Used in linking to /show-email
 SHOW_EMAIL_SECRET = os.environ.get('SHOW_EMAIL_SECRET', 'foo')
+
+# Used in linking to /ct
+CLICKTHROUGH_SECRET = os.environ.get('CLICKTHROUGH_SECRET', 'bar')
```
bdc34: There is no need for a separate clickthrough secrete key.
 app.config.get('SECRET_KEY') can be used.

tests/data/abs_files/ftp/acc-phys/papers/9411/9411001.abs
```diff
@@ -4,7 +4,7 @@ arXiv:acc-phys/9411001
 From: Salman Habib <NdENv@u0-4VU8.fWz4.XrM>
 Date: Tue, 1 Nov 1994 18:57:48 GMT   (199kb)
 
-Title: Symplectic Computation of Lyapunov Exponents
+Title: {\gamma}-Cascade Symplectic Computation of Lyapunov arXiv:1701.00001
```
bbarker: Was this `\gamma` actually present in the original abs? it might interfere with other tests if not. In general, curious why the abs file was changed

bbarker: Was this `\gamma` actually present in the original abs? it might interfere with other tests if not. In general, curious why the abs file was changed

mhl10: I committed this by accident; thanks for pointing it out. Will revert.

PR #30, opened 2018-09-26
ARXIVNG-1210: miscellaneous bugfixes

mhl10:
This PR addresses a number of bugs surfaced while testing 100K random papers on beta.arxiv.org in batch:

* handle the case where a versioned (not latest) .abs file does not include a "Categories:" field
* handle the case where submitter name is not provided
* change default encoding to latin-1 when opening .abs files
* add missing proxy section to version history display
* minor fix to author parsing

I added test data (.abs files) for all of these. This branch is currently active on beta.arxiv.org.


tests/data/abs_files/ftp/arxiv/papers/1310/1310.8286.abs
```diff
@@ -0,0 +1,95 @@
+------------------------------------------------------------------------------
+\\
+arXiv:1310.8286
+From: Brant M. Johnson <long_authors@example.org>
+Date: Wed, 30 Oct 2013 19:51:30 GMT   (242kb)
+
+Title: System-size dependence of open-heavy-flavor production in
+  nucleus-nucleus collisions at $\sqrt{s_{_{NN}}}$=200 GeV
+Authors: A. Adare, S. Afanasiev, C. Aidala, N.N. Ajitanand, Y. Akiba, H.
```
mhl10: This should probably be accompanied by another test.

browse/services/search/search_authors.py
```diff
@@ -131,7 +131,9 @@ def _link_for_name_or_collab(item: str) -> AuthorList:
         if re.search(r'Jr\b|Sr\b|[IV]{2, }]', name_bits[-1]):
             name_bits.pop()
 
-        surname = name_bits.pop()
+        surname = ''
```
mhl10: This may need some tweaking.

browse/domain/metadata.py
```diff
@@ -117,7 +117,7 @@ class DocMetadata:
     categories: str
```
bbarker: Should this (in DocMetadata) need to be Option[str] now to accommodate the possible `None` assignment that is added? Not sure how it typechecks, but I haven't tried running it locally yet.

browse/services/document/metadata.py
```diff
@@ -550,12 +565,13 @@ def parse_abs_file(filename: str) -> DocMetadata:
             abstract=fields['abstract'],
             authors=AuthorList(fields['authors']),
             submitter=Submitter(name=name, email=email),
-            categories=fields['categories'],
+            categories=fields['categories'] if 'categories' in fields else None,
```
bbarker: This is the line I am referring to in the other comment.

mhl10:
@bbarker the testing I'm referring to here is independent of the comparison script. The batch testing is meant to reveal error conditions in the form of HTTP exceptions (i.e. 500 errors). I have a simple one-liner that leverages `parallel` and `curl`; I'm running this against batches of 100K versioned IDs at a time, making requests like https://beta.arxiv.org/ng/abs/<idv>

browse/domain/metadata.py
mhl10: It seems to typecheck fine but I think `Optional[str]` would be clearer so I'll try that.

PR #37, opened 2018-10-11
HTML and minor bugfixes

mhl10:
Fixes:
* excessive spaces
* errant tags
* escaping
* browse context

### DIFF
--- a/browse/controllers/abs_page/__init__.py
+++ b/browse/controllers/abs_page/__init__.py
@@ -292,28 +292,26 @@ def _check_context(arxiv_identifier: Identifier,
     None
 
     """
+    context = None
     if 'context' not in request.args:
         if arxiv_identifier.is_old_id:
-            # This is just because legacy use the archive as the browse
+            # This is just because legacy uses the archive as the browse
             # context for old IDs.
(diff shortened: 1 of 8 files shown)


browse/controllers/abs_page/__init__.py
```diff
@@ -292,28 +292,26 @@ def _check_context(arxiv_identifier: Identifier,
...
     """
+    context = None
     if 'context' not in request.args:
         if arxiv_identifier.is_old_id:
-            # This is just because legacy use the archive as the browse
+            # This is just because legacy uses the archive as the browse
             # context for old IDs.
             response_data['browse_context'] \
                 = arxiv_identifier.archive
-    return
```
mhl10: this was the main problem with browse context

browse/templates/abs/trackback_rdf.html
```diff
@@ -1,6 +1,6 @@
+{#- TODO: check if dc:title is tex 2 utf8 in original system, need to wait til merge of PR for ARXIVNG-601 -#}
```
mhl10: turned this into a jinja comment so it wouldn't be included in the output

PR #40, opened 2018-10-16
Browse v0.1

mhl10:
This review is for Browse v0.1 (Milestone B01: Abstract Page). The primary purpose of the review is to bring the rest of the team up to speed with developments in browse.

The goal of this release is to reproduce the functionality, look and feel of the classic abstract ("abs") page, but with a new underlying architecture. With this release we are replacing the following two classic browse routes:
* /abs
* /ct (minor)

**Note**: other routes are not yet implemented and so clicking on some links may generate 404 responses when running the application locally.

At first glance the changeset will appear daunting, but a good chunk of this is in test files. Kindly complete your review before **COB on Friday, October 19**. Our goal is to deploy at the beginning of the week of October 22; any issues raised during this review will be addressed with patches.

Please address any questions to the release team:
@bdc34 
@mhl10 

# Getting Started
Please try running this release in your local development environment, following the instructions outlined in `README.md`.

Over 1,000 .abs files are included in the test data, in `tests/data/abs_files/`. You can enter the article identifier for any of these when running the application locally.

The release is also deployed on beta.arxiv.org. Here, the classic abs route has been moved to `/abs_classic`.

# How to review

## What to look for
- Clarity on code structure and ability to understand how the feature is built.
- Clarity on how this feature fits into documented architecture.
- Consistency of feature with decisions made in other parts of the project.
- Parity with current milestone: see milestone doc

## Feedback collection and format
- Code review: GitHub review comments
  - Inline comments are great to use for context on specific issues
  - “Approve” means “passes muster, ready to go as-is”
  - “Request changes” is exactly that; a bug or fix worth making now to avoid having obvious mistakes released to the wild
  - If neither option seems applicable: make comment to indicate that you have reviewed and are finished offering feedback
  - Avoid a non-response (no feedback at all), even if it’s just “I pass on this review and defer to others”

## Giving effective feedback
- Keep feedback within the scope of the target milestone.
- If something is unclear, assume that there is a good explanation and ask for clarification.
- As always, keep it civil, respectful, and supportive.

screenshot 2018-10-16 15 45 00

# Overview of Features

A summary of the features is available here: 
https://confluence.cornell.edu/display/arxivpub/Browse+v0.1

Some highlights are included below: 

## Extra Services
"Extra Services" refers to the functions in the right-hand box on the /abs page.

### Download Links
screenshot 2018-10-16 15 50 04

### Browse Context
screenshot 2018-10-16 15 52 06

**Note**: when run locally, the `prev` and `next` links will only work when viewing an "old" arXiv identifier, e.g. cs/0006008. New identifiers leverage the `/prevnext` route which has not yet been reimplemented. 

### Ancillary File Listing
screenshot 2018-10-16 16 07 30

**Note**: there are no source files included with the test data, so you will not see this section when running the app locally against the test data. See https://beta.arxiv.org/abs/0905.2326 or https://beta.arxiv.org/abs/1106.6351 for examples.

### DBLP Listing
screenshot 2018-10-16 15 47 34

e.g. cs/0006008 or 0704.0361 (both in test data)

### Trackback Count
screenshot 2018-10-17 16 13 58

e.g. 0808.4142 (in test data)

### References & Citations
screenshot 2018-10-16 15 50 49

**Note**: arXiv Bibliographic Explorer support is also included in the configuration and templates, but the js/css is not included in this project.

### Social Bookmarking
screenshot 2018-10-16 15 51 08

Links use the new clickthrough route (/ct).

## Changes from Classic
There are several minor changes from classic /abs page display:
* Submission history:
screenshot 2018-10-16 15 53 34

    * The submitter name shown on the "From:" line now reflects the submitter name from the version being viewed; classic only displayed the submitter name from the latest version
    * UTC is now used instead of GMT in the date line
    * The enigmatic, single-character format codes are now omitted from the date line; in classic these appeared after the source file size parenthetical
* Clickthrough (/ct) links:
    * The hash function being used to generate the clickthrough link hash value is different from classic, so the links are different, but the functionality is the same


browse/domain/metadata.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def __post_init__(self) -> None:
+        """Get the full group name."""
+        super().__post_init__()
+        if self.id in taxonomy.GROUPS:
+            self.name = taxonomy.GROUPS[self.id]['name']
+
+
+@dataclass(frozen=True)
+class DocMetadata:
```
mhl10: This is the main domain class for the document metadata.

browse/templates/abs/404.html
```diff
@@ -0,0 +1,88 @@
+{% extends "base.html" %}
+
+{%- block title -%}
+  {%- if reason -%}
+    {%- if (reason=='not_found' or reason=='version_not_found') and arxiv_id -%}
+[{{arxiv_id}}] Article does not exist
```
eawoods: If the reason for the 404 is not found -- can we say 'Article not found'? Is there ever a case in which our code might be in error and the article in question does exist but the path is wrong? A status of 'lost' is probably a lot less anxiety-inducing for an author than 'nonexistent', from a user-friendliness perspective.

browse/templates/abs/abs.html
```diff
@@ -0,0 +1,185 @@
...
+    <div class="authors"><span class="descriptor">Authors:</span>
+      {%- include "abs/author_links.html" %}
+    </div>
+
+    <div class="dateline">{{ generate_dateline() }}</div>
+
+    <blockquote class="abstract mathjax"><span class="descriptor">Abstract:</span>{{ abs_meta.abstract|arxiv_urlize|line_feed_to_br }}</blockquote>
+    <!--CONTEXT-->
+    <div class="metatable">
+      <table summary="Additional metadata">
```
eawoods: Technically this table is a layout table rather than a data table - can we, at some point, replace it with a more accessible layout? Does not need to block beta but bookmarking as a to-do that must get on the list.

browse/templates/abs/extra_services.html
```diff
@@ -0,0 +1,217 @@
...
+        (<a href="{{abs_meta.license.effective_uri}}" title="Rights to this article">license</a>)
+        {%- endif -%}
+      </div>
+    </div>
+    <!--end full-text-->
+
+{%- if ancillary_files %}
+  <div class="ancillary">
+    <span class="descriptor">Ancillary-file links:</span>
+    <h2>Ancillary files <span style="font-size:75%;font-weight:normal">(<a href="/src/0905.2326v2/anc">details</a>)</span>:</h2>
```
eawoods: hardcoded anc link to "details"?

browse/templates/base.html
```diff
@@ -0,0 +1,96 @@
...
+      </div>
+    </div>
+    <div id="support-ack">
+      <a href="{{url_for('acknowledgment')}}">We gratefully acknowledge support from<br/>the Simons Foundation<br /> and {{ session['institution'] if ('institution' in session and session['institution'] != None) else 'member institutions' }}</a>
+    </div>
+  </div>
+
+  <div id="header" {% block header_class %}{% endblock %}>
+    {% block header %}
+    <a href="{url_path('ignore_me')}"></a>
```
eawoods: Could we have a comment on this blank link saying what it's for? If it is visible to users/screen readers may need a class for it to explicitly be hidden or visible in some way.

browse/util/id_patterns.py
```diff
@@ -0,0 +1,323 @@
+r"""Patterns and functions to detect arXiv ids and Urls in text.
```
eawoods: random r at start of this line? or is it supposed to be there?

browse/util/id_patterns.py
```diff
@@ -0,0 +1,323 @@
...
+of hostnames to meet the needs of arxiv.
+
+HOST_NAME must end with a simplified character to avoid capturing a
+period.
+"""
+
+PATH = rf'(?P<PATH>(/{OKCHARS}*)+)?'
+"""Regex for path part of URLs for use in urlize"""
+
+QUERY = rf'(?P<QUUERY>\?(&?({OKCHARS}*(={OKCHARS}*)?))*)?'
```
eawoods: second QUERY has an extra U

browse/services/document/author_affil.py
```diff
@@ -0,0 +1,363 @@
+"""Parse Authors lines to extract author and affiliation data."""
+import re
+from itertools import dropwhile
+from typing import Dict, Iterator, List, Tuple
+
+from browse.services.util.tex2utf import tex2utf
```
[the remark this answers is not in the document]
mhl10: That's right, `tex2utf` is the function and `tex_to_utf` is the jinja filter that wraps around that function. We'll try to make that a bit clearer with comments in the code.

browse/templates/abs/404.html
mhl10: Good idea! I've replaced the "does not exist" language with "not found" in the `400.html` template.

browse/templates/base.html
mhl10: This is an old honeypot link that users ought to never see, but it will be good for us to check whether screen readers catch it. It's intended to trap robots/crawlers that are not respecting the contents of `robots.txt` (which indicates this link should not be crawled). I'll add a comment about it.

browse/util/id_patterns.py
bdc34: It is the flag to python that the string should be escaped for regexes.  It is there because the comment has slashes and one of the linters was complaining that it might be a regex.

mhl10: My bad, the linter did complain :-) . I've reverted my previous change.

DavidLFielding:
Overall looks great! I still need to perform rapid code review.

I've been looking at a local version of the Browse UI.

Two minor things caught my attention on new abs page. 

1) With tunnel vision enabled I temporarily thought I was looking at the wrong version of paper.

    I was clicking on different versions of arXiv:0906.5132v1 and noticed a slight difference, likely intentional, in the first version of paper.

(Submitted on 28 Jun 2009, latest version 8 Oct 2009 (v4))

    For a split second I thought I had version v4, published Oct 8th, Submitted July 28th. This is likely intentional but is this inconsistent with other version descriptions? Seems like (this version, v1) or (v1) would be closer to the other version description strings.

(Submitted on 28 Jun 2009 (v1), revised 21 Jul 2009 (this version, v2), latest version 8 Oct 2009 (v4))
(Submitted on 28 Jun 2009 (v1), revised 29 Jul 2009 (this version, v3), latest version 8 Oct 2009 (v4))
(Submitted on 28 Jun 2009 (v1), last revised 8 Oct 2009 (this version, v4))

2) DOI field ONLY appears for NON-versioned abstract page.

https://arxiv.org/abs/0906.5132

Subjects: | Superconductivity (cond-mat.supr-con); Materials Science (cond-mat.mtrl-sci)
DOI: | 10.1007/s10909-009-0032-7
Cite as: | arXiv:0906.5132 [cond-mat.supr-con]
  | (or arXiv:0906.5132v4 [cond-mat.supr-con] for this version)

https://arxiv.org/abs/0906.5132v1

Subjects: | Superconductivity (cond-mat.supr-con); Materials Science (cond-mat.mtrl-sci)
Cite as: | arXiv:0906.5132 [cond-mat.supr-con]
  | (or arXiv:0906.5132v1 [cond-mat.supr-con] for this version)

Again, this may be intentional, but seems a little odd to me.

I'll add any more UI issues I discover here. I still need to take a quick look at old identifiers...trying to get these to you as fast as possible.

bdc34:
I just checked. Yes the DOI behavior consistent with is what is happening on arxiv.org.

browse/services/database/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    MemberInstitution, MemberInstitutionIP, TrackbackPing, SciencewisePing, \
+    DBLP, DBLPAuthor, DBLPDocumentAuthor
+from arxiv.base import logging
+from logging import Logger
+
+logger = logging.getLogger(__name__)
+app_config = get_application_config()
+
+
+def db_handle_error(logger: Logger, default_return_val: Any) \
```
JaimieMurdock: Should something like this go into `arxiv.base`?

browse/services/document/author_affil.py
```diff
@@ -0,0 +1,363 @@
+"""Parse Authors lines to extract author and affiliation data."""
```
JaimieMurdock: should these go in `arxiv.base`? Seems of general use for arxiv-submission-ui and arxiv-search

browse/services/util/formats.py
```diff
@@ -0,0 +1,161 @@
...
+from tarfile import ReadError, CompressionError
+
+import logging
+logger = logging.getLogger(__name__)
+
+# List of tuples containing the valid source file name extensions and their
+# corresponding dissemintation formats.
+# There are minor performance implications in the ordering when doing
+# filesystem lookups, so the ordering here should be preserved.
+VALID_SOURCE_EXTENSIONS = [
```
JaimieMurdock: Also could be shared with `arxiv-compiler` so lift to `arxiv.base`?

browse/util/id_patterns.py
JaimieMurdock: DRY and `arxiv.identifier.parse_arxiv_id` - perhaps some of these functions should also be folded into [`arxiv.identifier` in arxiv-base](https://github.com/cul-it/arxiv-base/blob/master/arxiv/identifier/__init__.py)?

browse/domain/category.py
```diff
@@ -0,0 +1,81 @@
+"""Class that represents a single category."""
+
+from typing import Union, List
+from dataclasses import dataclass, field
+
+from arxiv import taxonomy
```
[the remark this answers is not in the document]
bdc34: I agree that much of this should go in arxiv-base. The ids and categories used by arXiv do not operate in obvious ways and it would be an unnecessary expense to duplicate them. 

Security through API keys is a better way to prevent the browse system from performing unauthorized actions on the file system. If we are confident that browse cannot access the file management system we don't need to care if it has the module on it's load path or not.

browse/services/database/__init__.py
bdc34: db_handler_error? It could if developers wanted it. Unlike arxiv IDs and categories, there currently isn't a plan to move utility code like this to arxiv-base.

browse/services/document/author_affil.py
bdc34: The plan for this was explicitly not to reuse the code in search for this. This decision could be revisited.

browse/services/util/formats.py
bdc34: Sounds good.

PR #47, opened 2018-11-07
Add rudimentary support for social media meta tags

mhl10:
We've learned from Steinn that ever since we modified the footer to include a11y information, the link previews (particularly those generated by FB) are sub-optimal. This PR is a first stab at including some of the social media meta tags per these [guidelines](https://css-tricks.com/essential-meta-tags-social-media/), which might help fix that display problem. Absent from this are the image tags that are technically required (but maybe, not really). 

I'm mainly putting this out there to continue the conversation, not necessarily to deploy something expediently (but that could be an option too).

### DIFF
--- a/browse/templates/abs/abs.html
+++ b/browse/templates/abs/abs.html
@@ -178,7 +178,7 @@
     <meta property="twitter:description" content="{{ abs_meta.abstract|trim|truncate(200, False, '...')|tex_to_utf }}"/>
     <meta property="og:site_name" content="arXiv.org"/>
     <meta property="og:title" content="{{ abs_meta.title|tex_to_utf }}"/>
-    <meta property="og:url" content="https://arxiv.org/{{ requested_id }}"/>
+    <meta property="og:url" content="{{ canonical_url(abs_meta.arxiv_id, abs_meta.version) }}"/>
     <meta property="og:description" content="{{ abs_meta.abstract|trim|tex_to_utf }}"/>
 {%- endmacro -%}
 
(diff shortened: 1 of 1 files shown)


browse/templates/abs/abs.html
```diff
@@ -171,6 +172,16 @@ <h2>Submission history</h2> From: {{ abs_meta.submitter.name|tex_to_utf if abs_m
 {%- endif -%}
 {%- endmacro -%}
 
+{%- macro generate_social_media_tags() -%}
+    <meta name="twitter:site" content="@arxiv"/>
+    <meta property="twitter:title" content="{{ abs_meta.title|tex_to_utf|truncate(70, False, '...') }}"/>
+    <meta property="twitter:description" content="{{ abs_meta.abstract|trim|truncate(200, False, '...')|tex_to_utf }}"/>
+    <meta property="og:site_name" content="arXiv.org"/>
+    <meta property="og:title" content="{{ abs_meta.title|tex_to_utf }}"/>
+    <meta property="og:url" content="https://arxiv.org/{{ requested_id }}"/>
```
erickpeirson: Is this right? Should this use ``url_for()`` so that the base server is take into account?

erickpeirson: By "is this right" I mean "what is requested_id and will this generate a usable url"

bdc34: How about `<meta property="og:url" content="{{ abs_meta.arxiv_identifier.canonical_url() }}"/>` ?

PR #48, opened 2018-11-13
ARXIVNG-1296 - Home page

mhl10:
This PR is for a simple reimplementation of the home page:
* uses DB to get the the document count by default; falls back to static stats file if DB unavailable. (Legacy defaulted to static stats file and never used DB)
* does **not** display login status (this will be a separate ticket)


browse/routes/ui.py
```diff
@@ -51,8 +61,8 @@ def bare_abs() -> Response:
     raise AbsNotFound
 
 
-@blueprint.route('/abs/', methods=['GET'], defaults={'arxiv_id': ''})
-@blueprint.route('/abs/<path:arxiv_id>', methods=['GET'])
+@blueprint.route('abs/', methods=['GET'], defaults={'arxiv_id': ''})
```
bdc34: Just curious, why is the leading slash being removed?

tests/test_browse.py
```diff
@@ -18,6 +19,17 @@ def setUp(self):
...
+        rv = self.app.get('/')
+        self.assertEqual(rv.status_code, 200)
+        html = BeautifulSoup(rv.data.decode('utf-8'), 'html.parser')
+
+        for group_key, group_value in taxonomy.GROUPS.items():
+            if group_key == 'grp_test':
+                continue
+            auths_elmt = html.find('h2', text=group_value['name'])
+            self.assertTrue(auths_elmt, f"{group_value['name']} in h2 element")
+
```
bdc34: In that continue on line 29, maybe add a test to make sure there are no groups named 'test'?

PR #53, opened 2018-12-17
ARXIVNG-1097 ARXIVNG-1388 ARXIVNG-1389 Abs macro move, and friends

erickpeirson:
See summary in https://github.com/arXiv/arxiv-base/pull/81 . I have all tests passing in browse at this point, hoping that's a good indication that I didn't completely break anything. I'm sure that there is more to do here before merging, but thought it was a good time to get another set of eyes.


browse/domain/category.py
```diff
@@ -6,76 +6,5 @@
...
-                return f'{arcname} ({self.id})'
-            else:
-                return self.id
-        else:
-            return self.id
-
-
-def _split_cat_str(cat: str)-> List[str]:
-    return cat.split('.', 2)
+class Category(taxonomy.Category):
```
erickpeirson: Might even get rid of this; just kept it as a placeholder

mhl10: I think it's OK to keep as a placeholder.

PR #56, opened 2019-01-10
/list page

bdc34:
This implements the /list controller and templates. 
Here are the types of paths that it implements:
/list/ARCHIVE/new
/list/ARCHIVE/YYMM
/listARCHIVE/YY
/list/ARCHIVE/recent
/list/ARCHIVE/pastweek (same as recent)
/list/ARCHIVE/current

It is currently using a very fake listing service. The same papers will be displayed for many combinations of ARCHIVE and time period.

https://arxiv-org.atlassian.net/browse/ARXIVNG-1291


browse/factory.py
```diff
@@ -32,12 +33,15 @@ def create_web_app() -> Flask:
     Base(app)
     app.register_blueprint(ui.blueprint)
 
+    #bdc34: not sure how to add service to app context
+    app.config['listing_service'] = FakeListingFilesService()
```
erickpeirson: You can use the Flask ``g`` object, see http://flask.pocoo.org/docs/1.0/appcontext/#storing-data.

browse/controllers/abs_page/__init__.py
```diff
@@ -70,7 +68,6 @@ def get_abs_page(arxiv_id: str) -> Response:
     ------
     :class:`.InternalServerError`
         Raised when there was an unexpected problem executing the query.
-
```
mhl10: Sorry to start off nitpicking, but my linter and pydocstyle will complain that this violates PEP257, D413 ([Missing blank line after last section](http://www.pydocstyle.org/en/2.1.1/error_codes.html)).

browse/services/document/author_affil.py
```diff
@@ -64,8 +64,7 @@ def parse_author_affil(authors: str) -> List[List[str]]:
 
 
 def _parse_author_affil_split(author_line: str) -> Dict:
-    """
-    Split author line into author and affiliation data.
+    """Split author line into author and affiliation data.
```
erickpeirson: Hmm, thought we got rid of this; should be using https://github.com/arXiv/arxiv-base/blob/master/arxiv/util/authors.py

browse/controllers/abs_page/__init__.py
mhl10: I guess this is not technically a problem with the numpy convention we're using.

browse/services/listing/__init__.py
```diff
@@ -0,0 +1,125 @@
...
+"""
+
+
+from datetime import date
+from typing import List, Optional, Tuple
+
+from mypy_extensions import TypedDict
+
+
+ListingItem = TypedDict('ListingItem',
```
erickpeirson: This and the other ``TypeDict`` below would probably be good candidates for the domain.

tests/test_list_page.py
```diff
@@ -0,0 +1,443 @@
+import unittest
+import re
+from hamcrest import *
```
erickpeirson: What assertions does hamcrest provide that are not already provided in the built-in unittest package?

browse/factory.py
bdc34: I changed this to use 'g'

tests/test_list_page.py
bdc34: I find the unittest.assertEqual() unusable due to its error messages. I use hamcrest to provide a message like "Expected 'Dec. 3rd'  but was  '12-03'"

browse/controllers/list_page/paging.py
```diff
@@ -0,0 +1,64 @@
+"""Paginiation links for listing pages."""
+
+
+from typing import Any, Dict, List, Optional, Tuple, cast
+import math
+from flask import current_app, request, url_for
+
+
+def paging(count: int, skipn: int, shown: int, context: str, subcontext: str) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
bdc34: I tried using a Union, it was easily doable and more precise.

browse/services/document/author_affil.py
mhl10: If the arxiv-base master branch is using it, then it ought to be in develop already

browse/controllers/list_page/__init__.py
```diff
@@ -0,0 +1,424 @@
...
+    'month': 'list/month.html',
+    'year': 'list/year.html'
+}
+
+
+def get_listing(
+        subject_or_category: str,
+        time_period: str,
+        skip: str,
+        show: str) -> Response:
```
mhl10: Curious why skip and show are of type `str`?

bdc34: They are str because that's what seemed to be getting passed from Flask.

browse/services/listing/__init__.py
mhl10: I agree

mhl10:
Looks good--very clear! Note the PR is still showing merge conflicts; it would be helpful to resolve those soon.

bdc34:
@mhl10  I just tried merging it and it didn't go smoothly. I'll try the merge again tomorrow.

browse/services/listing/__init__.py
bdc34: These are not dataclasses, they are just type data for the dict that the listing service returns. 
I'll move them if you guys want them in some other file than where they are used, otherwise I'm going to leave them here.

erickpeirson: Take a look at https://arxiv.github.io/arxiv-arxitecture/crosscutting/services.html#data-domain for a description of the role of the domain. The short answer is that if they are crossing the API of the service integration module inside of the browse service, then they are probably part of the domain. Domain classes can be dataclasses, NamedTuples, type aliases, etc.

browse/services/document/author_affil.py
bdc34: Deleted author_affil.py.

browse/services/listing/__init__.py
bdc34: Should this use just "from Flask import g" or "from arxiv.base.globals import get_application_global"?

PR #57, opened 2019-01-11
ARXIVNG-1393 - Reimplement /prevnext route

mhl10:
The `/prevnext` route is a navigational aid accessed via "< prev" and "next >" links on the `/abs` pages to navigate abstracts based on a subject-area context. In classic arXiv, this route was only used on documents with the new-style identifier; for the old-style identifiers, the links were generated by means of a filesystem lookup.

This implementation does away with the split-brain prev/next behavior from classic and now uniformly uses the /prevnext links. Behind the scenes, previous and next identifiers are determined via a DB lookup just like in classic, but now the old-style IDs are also supported.


browse/services/database/__init__.py
```diff
@@ -164,3 +168,51 @@ def get_document_count() -> Optional[int]:
...
     return document_count
+
+
+@db_handle_error(logger=logger, default_return_val=None)
+def get_sequential_id(paper_id: Identifier,
+                      context: str = 'all',
+                      is_next: bool = True) -> Optional[str]:
+    """Get the next or previous paper ID in sequence."""
+    local_session = db.session()
+    baked_query = bakery(lambda session: session.query(Document.paper_id))
```
mhl10: My use of the sqlalchemy bakery extension may be of interest to reviewers. I found this a useful mechanism for building database queries programmatically.

bdc34: Ya, that looks useful.

browse/services/database/models.py
```diff
@@ -293,6 +295,93 @@ class DBLPDocumentAuthor(db.Model):
     document = relationship('Document')
 
 
+class Category(db.Model):
```
bdc34: What is the plan for these categories in the DB and the taxonomy in arxiv-base?

browse/controllers/prevnext/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+                                   is_next=is_next,
+                                   context=context)
+        if seq_id:
+            redirect_url = url_for('browse.abstract',
+                                   arxiv_id=seq_id,
+                                   context=context)
+            return {},\
+                status.HTTP_301_MOVED_PERMANENTLY, {'Location': redirect_url}
+        else:
+            raise BadRequest(
```
bdc34: Should this be a 404 instead of a bad request?

browse/services/database/models.py
mhl10: None yet. For the time being they will have to coexist as before.

browse/controllers/prevnext/__init__.py
mhl10: Probably, and was wondering during reimplementation (classic uses same). It seemed borderline enough that I didn't get bogged down by it.

bdc34: Sounds fine either way. 404 seems more correct with HTTP semantics since the same next request might work in the future.

PR #60, opened 2019-01-23
Arxivng 1612 hotfix

bdc34:
The core of this problem is that in arxiv-browse, the abs template is using tex_2_utf on the abstract. This is converting TeX like /psi to the unicode φ. This is bad because φ doesn't look like a TeX /psi. 

This PR is step 1 and step 2 of plan B:
Step 1 In NG arxiv-browse have a new filter tex_2_utf_no_symbols
Step 2: In NG arxiv-browse, use that new filter on abstract only 

The changes to the legacy submission display are in the works.

The classic system does not do tex_2_utf, only normal HTML escaping: 
https://github.com/arXiv/arxiv-lib/blob/64ae85ad9168713198fc682ab3882cc5a3f02a21/lib/arXiv/HTML/Abstract.pm#L904

### DIFF
--- a/browse/factory.py
+++ b/browse/factory.py
@@ -62,7 +62,7 @@ def create_web_app() -> Flask:
 
     app.jinja_env.filters['line_feed_to_br'] = line_feed_to_br
     app.jinja_env.filters['tex_to_utf'] = partial( tex_to_utf, symbols=True )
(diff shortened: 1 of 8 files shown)


bdc34:
I created ARXIVNG-1623 to fix this bug in arxiv-base.

PR #61, opened 2019-01-28
Adding If-Modified-Since handling to the list controller.

bdc34:
The controller passes the If-Modified-Since value to the listing service and returns a 304 if the service response indicating that the listing hasn't changed.

### DIFF
--- a/browse/services/listing/__init__.py
+++ b/browse/services/listing/__init__.py
@@ -17,28 +17,31 @@ be better date granularity for new papers.
 """
 
 from typing import Optional
+from datetime import datetime
 
 from browse.domain.listing import NewResponse, ListingResponse
+
 
 class ListingService:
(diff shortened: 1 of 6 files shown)


browse/services/listing/__init__.py
```diff
@@ -17,28 +17,31 @@
...
 
 from browse.domain.listing import NewResponse, ListingResponse
 
+
 class ListingService:
     """Class for arXiv document listings."""
 
     @classmethod
-    def version(self) -> str:
+    def version(cls) -> str:
```
mhl10: I think the python convention is to do this:

`__version__ = '0.2'`

bdc34: I switched it from self to cls because the linter was complaining. This pattern is from ABC and allows subclasses to return superclass independent version values.

PR #63, opened 2019-02-08
ARXIVNG-1705: Hotfix to update footer

mhl10:
This is intended to be deployed with docs changes.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-arxiv-base = ">=0.9.1"
+arxiv-base = "==0.12.1rc2"
 dataclasses = "*"
 python-dateutil = "*"
 flask = "==0.12.*"
--- a/browse/templates/base.html
+++ b/browse/templates/base.html
@@ -93,6 +93,7 @@
     {% endblock content %}
   </div>
 
+  {#- TODO: url_for on as many links as possible -#}
   <footer>
     <div class="columns" role="navigation" aria-label="Secondary" style="margin: -0.75em -0.75em 0.75em -0.75em">
       <div class="column">
--- a/browse/templates/base.html
+++ b/browse/templates/base.html
@@ -128,15 +128,7 @@
       </div>
     </div>
   </footer>
-{#
-  <div id="footer">
-    <div class="footer-text">
-      <p>Link back to: <a href="{{url_for('home')}}">arXiv</a>, <a href="{{url_for('browse.form')}}">form interface</a>, <a href="{{url_for('contact')}}">contact</a>.&nbsp;&nbsp;</p>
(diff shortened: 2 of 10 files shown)


Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-arxiv-base = ">=0.9.1"
+arxiv-base = "==0.12.1rc2"
```
mhl10: Will need to fix some things before upgrading to latest version of arxiv-base.

browse/templates/base.html
```diff
@@ -92,13 +92,44 @@ <h1>arXiv.org</h1>
     {% block content %}
     {% endblock content %}
   </div>
-  <div id="footer">
-    <div class="footer-text">
-      <p>Link back to: <a href="{{url_for('home')}}">arXiv</a>, <a href="{{url_for('browse.form')}}">form interface</a>, <a href="{{url_for('contact')}}">contact</a>.&nbsp;&nbsp;<span class="help" style="display: inline-block; float: right"><a href="https://confluence.cornell.edu/x/MjmLFQ">Browse v0.1 released 2018-10-22</a>&nbsp;&nbsp;<button class="button is-small" id="feedback-button">Feedback?</button></span></p>
-      <p class="a11y-text" style="margin: 1em 0; padding-top: 1em; border-top: 1px solid #ccc">If you have a disability and are having trouble accessing information on this website or need materials in an alternate format, contact <a href="mailto:web-accessibility@cornell.edu">web-accessibility@cornell.edu</a> for assistance.</p>
+
+  {#- TODO: url_for on as many links as possible -#}
```
eawoods: arxiv-base has all of these links as url_for -- is it possible to copy config and code from there? Don't know how much trouble that would be.

eawoods:
@bdc34 the footer on the 404 page is different because that page is being served with an older version of arxiv-base (0.12.1rc2; current version is 0.14.3). I'd think having the 404 page for browse be slightly different than the rest of the site is OK until alignment can be made as per Martin's comment about upgrading. (if you need styling support for that, I'm available.)

browse/templates/base.html
```diff
@@ -35,6 +34,7 @@
   {%- include "feedback_collector_js.html" -%}
   {% if config['BROWSE_USER_BANNER_ENABLED'] %}
   <link rel="stylesheet" type="text/css" media="screen" href="//static.arxiv.org/css/user-banner.css?v=1.0" />
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css">
```
eawoods: this should probably come before the if statement for user_banner? otherwise it doesn't load on pages without user_banner.

mhl10: Good catch! As you can see I was in a hurry. Fixed!

browse/templates/base.html
mhl10: I'd like to upgrade to the latest version of arxiv-base in arxiv-browse while taking care of this TODO. I ran into some problems trying the upgrade before, so that's why I've pinned 0.12.1rc2 in `Pipfile` for now. I'll create a separate ticket/PR for the upgrade work.

PR #64, opened 2019-02-13
Reimplement /archive

bdc34:
Reproduces /archive route for browse.

To try it out:
clone the repo, check out the branch,
pipenv install
pipenv shell 
FLASK_APP=app.py FLASK_DEBUG=1 pipenv run flask run                           
Then go to some archives
Single category archive:
http://localhost:5000/archive/hep-ph
Multi category archive:
http://localhost:5000/archive/physics
Multi category archive with descriptions:
http://localhost:5000/archive/astro-ph
Subsumed category:
http://localhost:5000/archive/acc-phys


bdc34:
Note: there is a failing test that is also failing on develop. I'm currently ignoring that as it goes by in the test results but we could wait until it is fixed in develop.

browse/templates/archive/single_archive.html
```diff
@@ -0,0 +1,89 @@
...
+            {{ list_form.archive }}
+            {{ list_form.year}} {{ list_form.month}}
+            {{list_form.submit}}
+          </form>
+        </li>
+      </ul>
+    </li>
+    
+    <li><b>Catch-up:</b> 
+      <form action="TODO" method="post">
```
mhl10: Should this be going to legacy /catchup?

browse/templates/archive/single_archive.html
```diff
@@ -0,0 +1,89 @@
...
+    </li>
+    
+    <li><b>Catch-up:</b> 
+      <form action="TODO" method="post"> 
+        {{ catchup_form.archive }} 
+        Changes since: 
+        {{ catchup_form.day }} 
+        {{ catchup_form.month }} 
+        {{ catchup_form.year }}  
+        abstracts {{ catchup_form.abstracts }}
```
mhl10: In classic, the "abstracts" text appears after the form field.

browse/templates/archive/single_archive.html
bdc34: Yes.

browse/templates/archive/archive_list_all.html
```diff
@@ -0,0 +1,43 @@
+{%- extends "base.html" -%}
+
+{% block title %}Archives{% endblock %}
+
+{% block header_h1 %}
+<h1>> archives</h1>
```
mhl10: link to arXiv.org missing from the `header_h1` block:
screenshot 2019-02-26 14 55 31

(also happening on individual archive pages)

browse/templates/archive/single_archive.html
```diff
@@ -0,0 +1,89 @@
...
+{% block header_h1 %}
+<h1>
+  > <a href="{{url_for('.list_articles', context=archive_id, subcontext='recent')}}">{{archive_id}}</a>
+</h1>
+{% endblock %}
+
+
+{%- block content %}
+<div id='content'>
+  <h1>{{archive.name}} <span style="font-size: 70%;">(since {{archive.start_date}})</span></h1>
```
mhl10: Nitpicking here, but we should probably preserve the month granularity from classic in the display, since I think we're defaulting to the first day of each month for all the `start_date` values (mainly because we can't set an undefined day in `datetime`).

This would be fine:
``{{ archive.start_date.strftime('%B %Y') }}``

browse/templates/archive/archive_list_all.html
```diff
@@ -0,0 +1,43 @@
...
+
+{%- block content %}
+<div id='content'>
+  <h1>Archives</h1>
+
+  {% if bad_archive %}
+  <h1>No archive '{{bad_archive}}' -- available archives are:</h1>
+  {% endif %}
+  
+  <p>Choose one of the following archvies.</p>
```
mhl10: `archvies` -> `archives`

browse/controllers/archive_page/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+
+    if archive_id == "list":
+        return archive_index(archive_id, status=status.HTTP_200_OK)
+
+    archive = ARCHIVES.get(archive_id, None)
+    if not archive:
+        archive_id = CATEGORIES.get(archive_id, {}).get("in_archive", None)
+        archive = ARCHIVES.get(archive_id, None)
+        if not archive:
+            return archive_index(archive_id, status=status.HTTP_404_NOT_FOUND)
```
mhl10: It might be nice to include a message with this 404 at the top of the archive index, per classic.

`No archive 'blah' -- available archives are:`

browse/templates/archive/single_archive.html
bdc34: This should be fixed. The order of the yyyy mm and day too.

browse/controllers/archive_page/catchup_form.py
```diff
@@ -0,0 +1,54 @@
...
+    )
+    archive = HiddenField("archive", validators=[DataRequired()])
+    submit = SubmitField("Go")
+
+    def __init__(self, archive_id: str, archive: Dict[str, Any], years: List[int]):
+        super(CatchupForm, self).__init__()
+        self.year.choices = [(str(yer)[-2:], str(yer)) for yer in years]
+        self.archive.data = archive_id
+        # self.years = years
+        # TODO how many years are we suppose to show on the catchupform?
```
mhl10: The way you've done it looks fine to me. Old TODOs?

browse/templates/archive/archive_list_all.html
bdc34: This use to be handled by the base template. Should I just hardcode this?

browse/controllers/archive_page/catchup_form.py
bdc34: No, they are real todos. It only works because it is Feb.

bdc34:
> Looks great, with minor changes requested. You had mentioned a couple of "stray" sub-pages that we agreed shouldn't be part of this work. Do you remember what those were? Just want to record them somewhere.

Looks like:
 Catch up controller.
 Year controller.
  Find controller.

browse/controllers/archive_page/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+def archive_index(archive_id: str, status: int) -> Response:
+    """Landing page for when there is no archive specified."""
+    data: Dict[str, Any] = {}
+    data["bad_archive"] = archive_id
+
+    archives = [
+        (id, ARCHIVES[id]["name"])
+        for id in ARCHIVES.keys()
+        if id not in ARCHIVES_SUBSUMED and not id.startswith("test")
+    ]
```
bbarker: Just to confirm I know what is happening here, I think Python is flattening the elements of what is inside the `[` `]` to be a list. But it looks like this almost a reverse-map style of syntax, where the output is specified first (`(id, ARCHIVES[id]["name"])`) then the list of things to map over is specified (the filtered `id`s). Would be good to know what this is. Also, are `archives` just `List[str, str]`? might be nice to annotate to make it a little more clear

browse/controllers/archive_page/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+) -> List[Tuple[Any, str]]:
+    """Returns links to year pages."""
+    if not archive or not archive_id or not years:
+        return [("bogusURL", "NODATA")]
+    else:
+        return [(_year_stats_link(archive_id, i), str(i)) for i in years]
+
+
+def _year_stats_link(archive_id: str, num: int) -> Any:
+    return url_for(
```
bbarker: Does `url_for` return a string, or can it be treated as such? Might reduce many of the `Any`s found in the surrounding code. Mabe we could make our own `url_for` wrapper in arxiv-base to make it more clear if not. Sorry for all the nitpicks on this, and I know it is usually flask and friends at fault, but seeing `Any` always makes me wonder what is going on.

bdc34:
@bbarker 

> The archive pages look relatively simple enough that we probably don't need to repurpose our original /abs comparison script to be run on this, but I thought I'd throw it out there as an idea. The other aspect of it would be that maybe it doesn't matter as much if everything stays (almost) the same in /archive as it does in /abs.

I was thinking of doing something like that. I thought it would be useful to gather several thousands requests from the access logs and test them against NG /archive. It would also be good to test NG /list this way.

I was thinking that this could happen during the testing phase of this release.

browse/controllers/archive_page/__init__.py
bdc34: You have it. The reverse-map syntax is python list comprehension. 
I often use flatten (as in flatmap) to mean [[a],[b,c]] => [a,b,c] which isn't happening here. But I think you mean flaten in a metaphorical sense.

browse/controllers/archive_page/__init__.py
bdc34: I'm thinking that we could use more @no_type_check and file or block  # type: ignore ( https://github.com/python/mypy/issues/626 ).
In arXiv specific non-flask business logic code mypy might still be useful.

PR #65, opened 2019-02-14
Adding documentation

JaimieMurdock:
This commit adds the documentation for the module and the `update-docs.sh` script which will update the version at https://arxiv.github.io/arxiv-browse.

I noticed 2 code organization issues that I just wanted to confirm or flag for a refactor:
1. The `browse.controllers` module has submodules and might be better suited by just a flat structure - so instead of `home_page/__init__.py`, just use `home_page.py`.
2. `browse.util` doesn't have any code in it. Remove?

Other than that, let me know what you think.


JaimieMurdock:
I'm getting some failed unit tests, which is strange because the changes here only affect brand-new code for documentation. :confused:

bdc34:
Ya, there is a test that is failing in develop that we are aiming to fix ASAP.

bdc34:
Both of your suggestions sound good to me.

bdc34:
Jaimie, 
Why are there all these redundant seeming skelton .rst files? Do these need to be kept up-to-date by hand or are they generated from the source code?

Is the Makefile just to call through to sphinx?  Why?

The built pages look good.

JaimieMurdock:
I can't merge this branch because the coveralls status check never cleared. Would like help from an admin @mhl10 @erickpeirson

PR #67, opened 2019-02-19
ARXIVNG-1716: upgrade to latest version of arxiv-base and cleanup

mhl10:
This upgrades browse to the latest version of `arxiv-base` (currently set to reference the `develop` branch in git--to be updated to release in pypi later). 
Cleanup includes:
* reconciliation with changes in `master` branch (incl. new footer)
* removal of code, tests that have been migrated to `arxiv-base`


browse/util/id_patterns.py
```diff
@@ -0,0 +1,329 @@
+r"""Patterns and functions to detect arXiv ids and Urls in text.
+
```
erickpeirson: Why is this coming back?

mhl10: I think this came back with merge from master; will remove

PR #69, opened 2019-02-28
ARXIVNG-1942 Flask app should stick around a bit longer

erickpeirson:
We should test this out with Apache to make sure it works as expected.

From ARXIVNG-1942

After spending some time in Flask-SQLAlchemy: it’s a little bit worse than this, and I feel dumb for not spotting it sooner.

We are using the app factory pattern, so we have something like factory.py:

```python
def create_app() -> Flask:
   app = Flask(__name__)
   app.config....
   return app
```

and then we have wsgi.py

```python
def application(environ, start_response):
    for key, value in environ.items():
        os.environ[key] = value
    app = create_app()
    return app(environ, start_response)
```

the reason that we did this, from what I can see, is that we are trying to use Apache’s SetEnv directive to configure the app. And those key/value pairs only get passed via the request environ.

But it means that the app only lives just a microsecond longer than the request itself, and so the queue won’t live any longer than that.

Things we need to do:
- Make sure that the application is instantiated in the root of wsgi.py, not in the application callable.
- Use something other then os.environ to configure the Flask app when running on Apache.


mhl10:
Brian and I discussed this yesterday as part of a larger discussion in browse. We thought this looked good; we'll test and revisit this PR.

wsgi.py
```diff
@@ -4,9 +4,13 @@
...
+
+
 def application(environ, start_response):
     """WSGI application factory."""
+    # Apache passes config from SetEnv directives via the request environ.
     for key, value in environ.items():
-        os.environ[key] = str(value)
-    app = create_web_app()
-    return app(environ, start_response)
+        if key in _application.config:
```
erickpeirson: This might not be good enough. I ran into an issue where ``SERVER_NAME`` was getting passed in uwsgi, and it mucked things up. See what you think

bdc34:
@erickpeirson I'm looking into this and it seems:
1. We want to load config/secrets from OS env vars because that is what is going to work well with docker and kubernetes, std. practice etc.
2. We want to run multiple WSGI apps from a single apache: search and browse.
3. We are using mod_wsgi in apache to handle WSGI.
4. We are sending request time WSGI env vars from apache ng_flask.conf files to the apps to set config values such as DB password etc.
5. Setting OS env vars at the apache process level wouldn't work since search and browse need different values.

Currently (including the changes in this PR): 
1. This works on beta and production. (or we expect it to)
2. This is having a problem under uWsgi (in a docker container?)

Our goals:
1. We'd like to ensure that DB connection pool works. Or at a higher level, we'd like to know the maximum number of DB connections coming out of any flask app. We think that it doesn't work right now because we are seeing a lot of connections on production.
2. We'd like to have this work with OS env vars and whatever is needed to get it working in apache on beta/prod.

Does all that sound about right? What doesn't sound right?

erickpeirson:
> Setting OS env vars at the apache process level wouldn't work since search and browse need different values.

Yes, and even if we didn't IIRC those vars aren't inherited by the wsgi process.

> This works on beta and production. (or we expect it to)
> This is having a problem under uWsgi (in a docker container?)

This PR works in general on Apache and uWSGI; the only thing that you may want to double-check is that you actually want to set all of the keys that come through in ``environ`` and are in the config.

> We'd like to ensure that DB connection pool works. Or at a higher level, we'd like to know the maximum number of DB connections coming out of any flask app. We think that it doesn't work right now because we are seeing a lot of connections on production.

Yes. One contributing factor to why you're not seeing the desired pooling effect (might be the only reason) is that the app is getting torn down after each request, and thus the queuepool is GCed. This PR would make sure that the application instance gets created once, and is reused, which is the standard pattern.

bdc34:
:+1:

bdc34:
What do you think of running arxiv-browse and arxiv-search in their own uWSGI and then using apache with mod-uwsgi?
That would at least make things more uniform for configuration and env vars.

bdc34:
```import os

from browse.factory import create_web_app
from browse.services.database import models

def application(env, startResponse):
    startResponse("200 OK", [("Content-type", "text/plain")])

    os.environ["SQLALCHEMY_DATABASE_URI"] = 'mysql://arxivrw:xxxxxx4@localhost/arXiv'
    os.environ["SQLALCHEMY_POOL_SIZE"] = "10"
    os.environ["SQLALCHEMY_MAX_OVERFLOW"] = "0"

    poolis = b"none"
    app = create_web_app()
    with app.app_context():
        poolid = str(models.db.engine.pool).encode('UTF-8')
    
    rv = [
        b"pool id is " + poolid + b"\n" +
    ]
    return rv
```

bdc34:
Oh, I didn't mean to "close and comment" I meant to cancel.

bdc34:
With the wsgi.py as:
```from browse.factory import create_web_app
from browse.services.database import models

def application(env, startResponse):
    startResponse("200 OK", [("Content-type", "text/plain")])

    os.environ["SQLALCHEMY_DATABASE_URI"] = 'mysql://arxivrw:xxxxxx4@localhost/arXiv'
    os.environ["SQLALCHEMY_POOL_SIZE"] = "10"
    os.environ["SQLALCHEMY_MAX_OVERFLOW"] = "0"

    poolis = b"none"
    app = create_web_app() #app created every request
    with app.app_context():
        poolid = str(models.db.engine.pool).encode('UTF-8')
    
    rv = [
        b"pool id is " + poolid + b"\n" +
    ]
    return rv```

I get a page with:
pool id is <sqlalchemy.pool.impl.NullPool object at 0x7f375ffa22e8>
And the hex id changes on every request.

With the wsgi.py as:
```from browse.factory import create_web_app
from browse.services.database import models
app = create_web_app() #app created once on load of this code
def application(env, startResponse):
    startResponse("200 OK", [("Content-type", "text/plain")])

    os.environ["SQLALCHEMY_DATABASE_URI"] = 'mysql://arxivrw:xxxxxx4@localhost/arXiv'
    os.environ["SQLALCHEMY_POOL_SIZE"] = "10"
    os.environ["SQLALCHEMY_MAX_OVERFLOW"] = "0"

    poolis = b"none"
    with app.app_context():
        poolid = str(models.db.engine.pool).encode('UTF-8')
    
    rv = [
        b"pool id is " + poolid + b"\n" +
    ]
    return rv```

I get a page with:
pool id is <sqlalchemy.pool.impl.NullPool object at 0x7f375ffa22e8>
And the hex id stays the same on every request.

I think this is an improvement, but I doubt if a NullPool does anything useful.

**Comment by erickpeirson** [comment]
> I think this is an improvement, but I doubt if a NullPool does anything useful.

You might be getting a NullPool because ``SQLALCHEMY_POOL_SIZE`` is not getting set in the config: https://github.com/arXiv/arxiv-browse/blob/develop/browse/config.py#L213 . IIRC NullPool is . only used when the pool size is set to 0.

**Comment by bdc34** [comment]
:+1:  I'll give that a try.

**Comment by bdc34** [comment]
Yep, I was configuring the pool size on the os.environ instead of the app.config. I'm now getting a stable 
```
pool id is <sqlalchemy.pool.impl.QueuePool object at 0x7f375ff754e0>
```
I'm going to check from the mysql side that the connections seem to be reused and the number of connections is limited.

**Comment by bdc34** [comment]
The mysql side that SQLAlchemy connection pooling looks good with a simple setup with arxiv-browse.services.database.model. It's servicing multiple threads, reusing mysql connections, reconnecting to a restarted db and limiting the max number of connections. 

A SQLAlchemy connection pool is shared across threads. The wsgi deamon process setting for thread count should be equal to the number of SQLAlchemy connection pool size. It doesn't need to be larger. It should not be smaller because if there are the maximum number of threads, there will not be enough connections to guarantee connections for each threads of them. 

This does create a working connection pool (which is what this PR is doing):
```    
app = create_web_app() # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< app outside of application()
def application(env, startResponse):
    startResponse("200 OK", [("Content-type", "text/plain")])
    # SQLAlchemy configuration on flask app
    app.config["SQLALCHEMY_DATABASE_URI"] = 'mysql://arxivrw:xxxxxxx@localhost/arXiv'
    app.config["SQLALCHEMY_POOL_SIZE"] = 3
    app.config["SQLALCHEMY_MAX_OVERFLOW"] = 0

    poolis = b"none"
    count = b"not set"
    with app.app_context():
        poolid = str(models.db.engine.pool).encode('UTF-8')
...
```

This does **not** create a working connection pool:
```    
def application(env, startResponse):
    startResponse("200 OK", [("Content-type", "text/plain")])
    app = create_web_app() # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< app inside of application()
    # SQLAlchemy configuration on flask app
    app.config["SQLALCHEMY_DATABASE_URI"] = 'mysql://arxivrw:xxxxxxx@localhost/arXiv'
    app.config["SQLALCHEMY_POOL_SIZE"] = 3
    app.config["SQLALCHEMY_MAX_OVERFLOW"] = 0

    poolis = b"none"
    count = b"not set"
    with app.app_context():
        poolid = str(models.db.engine.pool).encode('UTF-8')
...
```

erickpeirson:
Glad it's having the desired effect. @mhl10 I will leave this open until you've had a chance to look, esp since it has implications for the Apache config

PR #70, opened 2019-03-01
ARXIVNG-1918 add four column footer

eawoods:
This implementation only affects browse pages (front, list, category, abs, etc).
Will have a separate implementation for NG-base shortly.

Note: right now the list and front pages don't have the extra space. I chose to not mess with it for now due to other possible layout complications.

## Desktop:
screen shot 2019-03-01 at 11 48 13 am

## Tablet:
screen shot 2019-03-01 at 11 48 30 am

## Mobile:
screen shot 2019-03-01 at 11 48 43 am

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,7 +16,7 @@ mysqlclient = "*"
 "mmh3" = "*"
 aiohttp = "*"
 flask = "==1.0.2"
-arxiv-base = "==0.15.1rc1"
+arxiv-base = {editable = true,git = "https://github.com/arXiv/arxiv-base.git",ref = "develop"}
 validators = "*"
 mypy-extensions = "*"
 flask-wtf = "*"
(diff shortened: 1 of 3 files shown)


erickpeirson:
@eawoods When I view this locally, I see the four columns, but the spacing relative to the abs content looks the same as what is on production now. Is that intended?

image
image

eawoods:
No - local is not pulling in the new static style sheet changes. Sorry about that - fixing now.

eawoods:
OK @erickpeirson try again - fixed style sheet and took out some padding that was causing issues. Will update screenshots momentarily.

mhl10:
I've updated dependencies in `Pipfile` which will hopefully fix failed tests.

mhl10:
Merging on Liz's behalf per discussion

PR #71, opened 2019-03-01
ARXIVNG-1356: auth integration with login status in header

mhl10:
Also includes ARXIVNG-2027.


browse/templates/home/home.html
```diff
@@ -28,11 +28,20 @@
...
 {#- TODO: define display order in taxonomy? -#}
 {{- group_section(('grp_physics','grp_math','grp_cs','grp_q-bio','grp_q-fin','grp_stat','grp_eess','grp_econ')) }}
-{% endblock content -%}
 
-{%- block footer_content -%}
-{%- include "long_trailer.html" -%}
-{%- endblock footer_content -%}
+<hr />
+{#- TODO: remove because of new footer? -#}
+<h2>About arXiv</h2>
```
mhl10: I forgot about this section of the homepage originally. I wonder if it should be removed or revised in light of new footer.

browse/config.py
```diff
@@ -255,3 +255,27 @@
...
+AUTH_UPDATED_SESSION_REF = True
+
+CLASSIC_COOKIE_NAME = os.environ.get('CLASSIC_COOKIE_NAME', 'tapir_session')
+CLASSIC_PERMANENT_COOKIE_NAME = os.environ.get(
+    'CLASSIC_PERMANENT_COOKIE_NAME',
+    'tapir_permanent'
+)
+CLASSIC_TRACKING_COOKIE = os.environ.get('CLASSIC_TRACKING_COOKIE', 'browser')
+CLASSIC_DATABASE_URI = os.environ.get('CLASSIC_DATABASE_URI', os.environ.get(
+    'BROWSE_SQLALCHEMY_DATABASE_URI', 'sqlite:///../tests/data/browse.db'))
```
bdc34: This looks like configuration related to testing. It seems like that should be setup in the tests, not in the config?
Failure to correctly configure the env vars would result in a production system running a test db.

browse/templates/login.html
```diff
@@ -0,0 +1,7 @@
+<div class="login">
```
bdc34: Could this be moved to auth?

browse/config.py
mhl10: This is configuration related to auth. Some options may not be needed for browse purposes. I'll review.

browse/templates/login.html
mhl10: I'm not sure what you mean. The template, or something else?

bdc34: The template for the login could live in auth. So it would be consistent across browse, search, auth, mod, admin, labs, etc.

bdc34:
I didn't notice that this was a "work in progress" PR. Sorry about commenting extra early.

browse/templates/home/home.html
bdc34: I'd leave it for now. Some other time there will be a larger scale redesign and it can get done during that.

browse/config.py
bdc34: I my concern here is that a missing configuration will spin up this code with a test DB. I'd prefer that missing configuration die with "missing CLASSIC_DB_URI" That might not be the model we are using though.

mhl10: What do you think about this tweak:

`CLASSIC_DATABASE_URI = os.environ.get('CLASSIC_DATABASE_URI', os.environ.get(
    'BROWSE_SQLALCHEMY_DATABASE_URI', None))`

Generally this should never hit the `None` case, but if it does, we could raise an error that prevents the app from starting. In the case where it's fallen back to sqlite, we could issue a warning.

bdc34: :+1:

PR #73, opened 2019-03-07
ARXIVNG-1968: four column footer

mhl10:


browse/templates/base.html
```diff
@@ -6,7 +6,7 @@
   {%- block head -%}
   <title>{% block title %}{% endblock %}</title>
   <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" />
-  <link rel="stylesheet" type="text/css" media="screen" href="//static.arxiv.org/css/arXiv-20190208.css?v=20190208" />
+  <link rel="stylesheet" type="text/css" media="screen" href="//static.arxiv.org/css/arXiv-ARXIVNG-1918.css?v=20190307" />
```
erickpeirson: Is ``arXiv-ARXIVNG-1918.css`` right? Does it exist?

mhl10: Thanks for pointing this out. It does exist. Once I deploy browse and confirm all is OK, I'll update the "master" copy arXiv.css in the CDN and point browse to that. Want to minimize issues due to caching.

browse/templates/base.html
```diff
@@ -38,10 +38,17 @@
   <script language="javascript" src="//code.jquery.com/jquery-latest.min.js"></script>
   <script language="javascript" src="//static.arxiv.org/js/user_banner.js"></script>
   {% endif -%}
+  <style type="text/css">
+    @media screen and (max-width: 1023px) {
```
eawoods: hold for update to the width on this number, I have to reconcile this with other pages (it forces columns on footer)

eawoods: yep, this rule that I added should now be removed. It's superseded in the arxiv-ARXIVNG-1918.css stylesheet.

PR #79, opened 2019-04-05
Changes how WSGI keeps flask app around

bdc34:
The main difference between this and the first branch of this name is that this one will copy all string values from wsgi.app() environ to os.environ. This will allow SetEnv directives in apache HTTPD conf files get all the way to values in browse.config.

In the first branch of this name wsgi.app() was using the keys of browse.config as a guide to which keys to copy from wsgi.app() environ. This ended up missing BROWSE_SQLALCHEMY_DATABASE_URI and some others.

Other changes:
The flask app is kept around between HTTP requests in a module level binding. Please check if the use of 'global' makes sense.

Changes to the documentation in browse/config.py. I moved some comments to doc strings.


wsgi.py
```diff
@@ -1,16 +1,53 @@
...
+
+    # 'global' actually means module scope, and that is exactly what
+    # we want here.
+    #    
+    # Python docs are thin. I'm seeing this sort of thing on
+    # stackoverflow: "In Python there is no such thing as absolute
+    # globals automatically defined across all namespaces
+    # (thankfully). As you correctly pointed out, a global is bound to
+    # a namespace within a module..."    
+    global __flask_app__
```
bdc34: @erickpeirson Please take a look at this global and let me know if it makes sense or of there is a better way to do this.

erickpeirson: Yep, looks good

wsgi.py
```diff
@@ -1,16 +1,53 @@
...
+    environ agrument. 
+
+    This will not be needed once each app is on docker+enginx.
+    """
+
+    # Copy string WSGI envrion to os.environ. This is to get apache
+    # SetEnv vars.  It needs to be done before the call to
+    # create_web_app() due to how config is setup from os in
+    # browse/config.py.
+    for key,value in environ.items():
```
erickpeirson:
```suggestion
    for key, value in environ.items():        
```

wsgi.py
```diff
@@ -1,16 +1,53 @@
...
+    This will not be needed once each app is on docker+enginx.
+    """
+
+    # Copy string WSGI envrion to os.environ. This is to get apache
+    # SetEnv vars.  It needs to be done before the call to
+    # create_web_app() due to how config is setup from os in
+    # browse/config.py.
+    for key,value in environ.items():        
+        if type(value) is str:
+            os.environ[key]=value
```
erickpeirson:
```suggestion
            os.environ[key] = value
```

PR #80, opened 2019-04-08
Browse v0.2.1

mhl10:
This review is for Browse v0.2.1 (Milestone B02: Additional Browse Routes). The primary purpose of the review is to bring the rest of the team up to speed with developments in browse.

The goal of this release is to reproduce the functionality, look and feel of additional browse routes, but with the new underlying architecture. It also lays the groundwork for a listing service, described more below.

With this release we are replacing the following classic browse routes:
* home page (`/`, `/index`), with login status via [arxiv-auth](https://github.com/arXiv/arxiv-auth) integration
* archive (`/archive/<archive_id>`)
* cookies (`/cookies`)
* stats (`/stats/today`, `/stats/monthly_submissions`, `/stats/monthly_downloads`)
* trackbacks (`/tb/<article_id>`, `/tb/recent`)
* previous/next navigation (`/prevnext`) 

The following routes are essentially complete but depend on a mock listing service that will be completed as part of the next browse milestone:
* list (`/list`)
* year (`/year`)

**We request that feedback be provided by COB Friday, April 12.**

Please address any questions to the release team:
@bdc34 
@mhl10 

# Getting Started
You can try running this release in your local development environment, following the instructions outlined in `README.md`.

The release is also deployed on beta.arxiv.org.  It may be more useful and expedient to test functionality (like "prev" and "next" navigation) here, since the test data is more comprehensive.

# How to review

## What to look for
- Clarity on code structure and ability to understand how the feature is built.
- Clarity on how this feature fits into documented architecture.
- Consistency of feature with decisions made in other parts of the project.
- Parity with current milestone: see milestone doc

## Feedback collection and format
- Code review: GitHub review comments
  - Inline comments are great to use for context on specific issues
  - “Approve” means “passes muster, ready to go as-is”
  - “Request changes” is exactly that; a bug or fix worth making now to avoid having obvious mistakes released to the wild
  - If neither option seems applicable: make comment to indicate that you have reviewed and are finished offering feedback
  - Avoid a non-response (no feedback at all), even if it’s just “I pass on this review and defer to others”

## Giving effective feedback
- Keep feedback within the scope of the target milestone.
- If something is unclear, assume that there is a good explanation and ask for clarification.
- As always, keep it civil, respectful, and supportive.

# Overview of Features
A summary of the features is available here: 
https://confluence.cornell.edu/display/arxivpub/Browse+v0.2

Some highlights are included below: 
## Home page (`/`, `/index`):
* supports paper count via DB query with fallback to classic, file-based method
* displays login status (best to test on beta)
Screenshot 2019-04-08 09 26 49

## Archive pages (`/archive/<archive_id>`):
Screenshot 2019-04-08 10 08 29

## Cookies (`/cookies`):
Screenshot 2019-04-08 10 01 59

### article trackbacks (`/tb/<article_id`)
* near-identical conversion with minor styling changes
Screenshot 2019-04-08 09 59 35

### recent trackbacks (`/tb/recent`)
* near-identical conversion with minor styling changes
Screenshot 2019-04-08 09 45 32

### today's stats (`/stats/today`)
Screenshot 2019-04-08 10 02 34

### monthly submissions (`/stats/monthly_submissions`)
Screenshot 2019-04-08 10 03 05

### monthly downloads (`/stats/monthly_downloads`)
Screenshot 2019-04-08 10 03 55

### previous/next abs navigation (`/prevnext`)
With the browse release 0.1 we recreated the hybrid behavior of arXiv classic "prev" / "next"; i.e. depending on the arXiv identifier being viewed (old or new), the system would perform a filesystem seek or database query to calculate the next or previous ID. This release switches everything to the database query method.
Screenshot 2019-04-08 14 01 36

## Pages that depend on list service
Controllers for the following routes have been implemented but will not be deployed as user-facing because they currently depend on a mock listing service.
### list (`/list/<archive_id>`)
Screenshot 2019-04-08 13 54 39

### year (`/year/<archive_id>`)
Screenshot 2019-04-08 13 55 03


mhl10:
@erickpeirson we're getting an exception on beta in the duplicate cookie scenario, specifically when it's hitting the redirect in `detect_and_clobber_dupe_cookies()`:

https://github.com/arXiv/arxiv-auth/blob/ec0f9c37a4bf9dce6f3eb51f9d98081302e9355a/users/arxiv/users/auth/__init__.py#L165

The `ui.login` route is defined in the accounts app outside the arxiv-auth package, so this may need to be changed unless there's a way to get the blueprints from accounts registered in browse? I might just be missing something, though!

mhl10:
> Overall looks great!
> One small thing: when moving from /archive to /list routes, the font size changes significantly (12.96px to 14.4px). Looks like the /archive pages have two nested `<div id="content">` wrappers, which is knocking the font size down twice.

Good catch! Fixed.

browse/controllers/archive_page/by_month_form.py
```diff
@@ -0,0 +1,49 @@
+"""Form for month selection of list controller."""
+from typing import List, Any, Dict
+
+from flask_wtf import FlaskForm
+from wtforms import SelectField, SubmitField, HiddenField
+from wtforms.validators import DataRequired
+
+MONTHS = [
```
erickpeirson: See above

browse/controllers/archive_page/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _write_expires_header(response_headers: Dict[str, Any]) -> None:
+    """Writes an expires header for the response."""
+    response_headers["Expires"] = abs_expires_header()[1]
+
+    
+DAYS = ["{:0>2d}".format(i) for i in range(1, 32)]
+
+MONTHS = [
```
erickpeirson: Any reason not to define this programmatically, using https://docs.python.org/3/library/calendar.html#calendar.month_abbr ?

```python
MONTHS = [(m, '%s (%s)' % (m, calendar.month_abbr[i])) for i, m 
          in enumerate(map(lambda o: str(o).zfill(2), range(0, 13) ))][1:]
```

browse/controllers/list_page/paging.py
```diff
@@ -0,0 +1,65 @@
+"""Paginiation links for listing pages."""
+
+
+from typing import Any, Dict, List, Union
+import math
+from flask import url_for
+
+
+def paging(count: int, skipn: int, shown: int, context: str, subcontext: str) \
+    -> List[Dict[str, Union[str, int]]]:
```
erickpeirson:
```suggestion
        -> List[Dict[str, Union[str, int]]]:
```

browse/controllers/list_page/paging.py
```diff
@@ -0,0 +1,65 @@
...
+    """Get paging links."""
+    B = 3  # num of buffer pages on each side of current
```
erickpeirson: Unless these are widely recognized/conventional symbols, would prefer to avoid single-character names. A quick glance suggests that it would not be too unwieldy to call this ``number_of_buffer_pages``, for example, which would increase legibility.

browse/controllers/list_page/paging.py
```diff
@@ -0,0 +1,65 @@
...
+    L = math.floor(count-1 / (skipn+1))+1  # total number of pages
+    S = 2 * B + 2 * 2 + 1  # number of total links in the pages sections:
+    #  2*buffer + 2*(first number + dots) + current
+
+    def page_dict(n: int, nolink: bool = False)->Dict[str, Union[str, int]]:
+        txt = str(n+1)+'-'+str(min(count, n+shown))
```
erickpeirson: [PEP008 recommends](https://www.python.org/dev/peps/pep-0008/#other-recommendations) whitespace around operators. Also, consider an F-string for readability.
```suggestion
        txt = f'{n + 1}-{min(count, n + shown)}'
```

browse/controllers/prevnext/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+            raise BadRequest('Missing or invalid function request')
+
+        if 'context' not in request_params:
+            raise BadRequest('Missing context')
+        context = request_params['context']
+
+        if not (context in CATEGORIES_ACTIVE
+                or context in ARCHIVES or context == 'all'):
+            raise BadRequest('Invalid context')
+    except IdentifierException:
```
erickpeirson: This is a pretty big ``try`` block; consider targeting the line that is liable to raise the ``IdentifierException``. Also, be sure to use [exception chaining](https://www.python.org/dev/peps/pep-3134/).

```python
try:
    arxiv_id = Identifier(request_params['id'])
except IdentifierException as e:
    raise BadRequest(f"Invalid article identifier {request_params['id']}") from e .   # <- chained exceptions
```

browse/controllers/prevnext/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+            return {},\
+                status.HTTP_301_MOVED_PERMANENTLY, {'Location': redirect_url}
+        else:
+            raise BadRequest(
+                f'No {"next" if is_next else "previous"} article found for '
+                f'{arxiv_id.id} in {context}'
+            )
+    except BadRequest:
+        raise
+    except Exception as ex:
```
erickpeirson: Any reason to catch these unhandled exceptions here beyond logging? Creates a lot of code overhead, especially catching your own raised ``BadRequest`` (L79 --> L83).

browse/controllers/stats_page/__init__.py
```diff
@@ -0,0 +1,144 @@
...
+        normal_count, admin_count, num_nodes = \
+            get_hourly_stats_count(stats_date=requested_dt.date())
+
+        response_data['current_dt'] = current_dt
+        response_data['requested_dt'] = requested_dt
+        response_data['normal_count'] = normal_count
+        response_data['admin_count'] = admin_count
+        response_data['num_nodes'] = num_nodes
+        return response_data, status.HTTP_200_OK, {}
+    except (TypeError, ValueError):
```
erickpeirson: Again, big ``try`` blocks -- it makes it a little harder to see where you expect the problem to arise, and what the nature of these exceptions might be. And don't forget chaining.

browse/controllers/year.py
```diff
@@ -0,0 +1,105 @@
...
+
+    return response_data, response_status, response_headers
+
+
+ASCII_ART_STEP = 20
+ASCII_ART_CHR = '|'
+ASCII_ART_URL_STEP = 100
+
+
+def ascii_art_month(archive_id: str, month: MonthCount) -> List[Tuple[str, Optional[str]]]:
```
erickpeirson: ASCII art makes me happy

browse/domain/listing.py
```diff
@@ -0,0 +1,125 @@
...
+from typing import List, Tuple
+
+from mypy_extensions import TypedDict
+
+
+ListingItem = TypedDict('ListingItem',
+                        {'id': str,
+                         'listingType': str,
+                         'primary': str})
+'''A single item for a listing.
```
erickpeirson: Minor nit: we're using double-quotes for docstrings everywhere else

browse/factory.py
```diff
@@ -1,79 +1,54 @@
...
-    models.init_app(app)
+    models.init_app(app)  # type: ignore
 
     Base(app)
+    Auth(app)
     app.register_blueprint(ui.blueprint)
 
-    ct_url_for = partial(create_ct_url, app.config.get(
-        'CLICKTHROUGH_SECRET'), url_for)
+    app.config['listing_service'] = FakeListingFilesService()
```
erickpeirson: This is not a pattern that we want to use. I realize that this is a fake service, but let's not let this stick around. The ``config`` is for configuration parameters. Instances of services can be bound to the appropriate context using the [``g`` object](http://flask.pocoo.org/docs/1.0/appcontext/#storing-data).

browse/routes/ui.py
```diff
@@ -64,97 +83,201 @@ def abstract(arxiv_id: str) -> Response:
...
+def list_articles(context: str, subcontext: str) -> Response:
+    """List articles by context, month etc.
 
-    Context might be a context or an archive
-    Subcontext should be 'recent' 'new' or a string of format yymm
+    Context might be a context or an archive Subcontext should be
+    'recent' 'new' or a string of format yymm
     """
-    raise InternalServerError(f'Not yet implemented {current_context} {yymm}')
+    if request.args.get('archive', None) is not None:
```
erickpeirson: L175-L181 are looking a lot like controller logic. Can we move this to ``list_page.get_listing()``?

browse/routes/ui.py
```diff
@@ -164,25 +287,68 @@ def show_email(show_email_hash: str, arxiv_id: str) -> Response:
...
+    """Landing page for an archive."""
+    response, code, headers = archive_page.get_archive(archive)  # type: ignore
+    if code == status.HTTP_200_OK or code == status.HTTP_404_NOT_FOUND:
+        return render_template(response['template'], **response), code, headers
+    elif code == status.HTTP_301_MOVED_PERMANENTLY:
+        return redirect(headers['Location'], code=code)
+    elif code == status.HTTP_304_NOT_MODIFIED:
+        return '', code, headers
+    else:
+        return response, code, headers
```
erickpeirson: Minor nit: dropping this out of the ``else`` block makes it clearer that this is the catch-call path. Looks like you guys are doing this in some cases, e.g. L338, below.

browse/services/database/__init__.py
```diff
@@ -155,3 +205,147 @@ def get_dblp_authors(paper_id: str) -> List[str]:
...
+    if paper_id.is_old_id:
+        # NB: classic did not support old identifiers in prevnext
+        if context == 'all':
+            like_id = f'{paper_id.archive}/{paper_id.yymm}%'
+        else:
+            like_id = f'%/{paper_id.yymm}%'
+    else:
+        like_id = f'{paper_id.yymm}.%'
+
+    baked_query += lambda q: q.filter(
```
erickpeirson: I realize that the SQLAlchemy docs shows examples with lambdas like this, too, but it sure smells a lot like assigning a lambda (see https://www.python.org/dev/peps/pep-0008/#programming-recommendations)... any reason not to explicitly define these?

browse/controllers/archive_page/by_month_form.py
erickpeirson: Or, below I guess.

browse/services/database/__init__.py
JaimieMurdock: The `BakedQuery` extension is the root of this construction: https://docs.sqlalchemy.org/en/13/orm/extensions/baked.html

The primary motive for using this system is for performance increases when composing a query. It would be possible to edit the `Query` object iteratively like so:
```
query = Session.query(Document.paper_id)
query = query.filter(Document.paper_id.like(bindparam('like_id'))
```

However, by using the `baked()` construction, we get the added benefit of a query cache, speeding up subsequent calls to the query.

One question that remains though is whether, because the queries in each of these functions is executed once per request, whether we're introducing overhead for no benefit at all -- if a new baked query is getting constructed on each function call to `services.database`, then the cache isn't doing anything and clarity of code might be the bigger advantage.

browse/templates/feedback_collector_js.html
```diff
@@ -8,8 +8,8 @@
     });
   },
   fieldValues: {
-    "components": ["15700"],  // Browse component.
-    "versions": ["14132"],  // Release browse-0.1
+    "components": ["15700"],  // browse component
```
JaimieMurdock: These are "magic numbers" from JIRA, right? might be worth highlighting where they came from.

browse/controllers/archive_page/__init__.py
bdc34: I would not choose to define it programmatically because the literal syntax seems simpler and more clear as to the final shape of the data.

erickpeirson: That's fair

bdc34:
The crosses are in a different order for https://beta.arxiv.org/abs/0906.3421 compared to https://beta.arxiv.org/abs_classic/0906.3421

Classic:  Statistical Mechanics (cond-mat.stat-mech); Mathematical Physics (math-ph)
NG: Mathematical Physics (math-ph); Statistical Mechanics (cond-mat.stat-mech)

Is this a bug or expected behavior?

mhl10:
> The crosses are in a different order for https://beta.arxiv.org/abs/0906.3421 compared to https://beta.arxiv.org/abs_classic/0906.3421
> 
> Classic: Statistical Mechanics (cond-mat.stat-mech); Mathematical Physics (math-ph)
> NG: Mathematical Physics (math-ph); Statistical Mechanics (cond-mat.stat-mech)
> 
> Is this a bug or expected behavior?

I thought secondaries were supposed to be sorted alphabetically by category ID rather than category description, so probably a bug? In any case, the sorting should be deterministic.

bdc34:
> I thought secondaries were supposed to be sorted alphabetically by category ID rather than category description, so probably a bug? In any case, the sorting should be deterministic.

It does seem deterministic. And I think you are right, they look like they are getting sorted by category description.  
Here is an issue for this: https://arxiv-org.atlassian.net/browse/ARXIVNG-2066

mhl10:
> > I thought secondaries were supposed to be sorted alphabetically by category ID rather than category description, so probably a bug? In any case, the sorting should be deterministic.
> 
> It does seem deterministic. And I think you are right, they look like they are getting sorted by category description.
> Here is an issue for this: https://arxiv-org.atlassian.net/browse/ARXIVNG-2066

It looks like the earlier NG browse release sorts this correctly, so this appears to be a regression:
https://arxiv.org/abs/0906.3421

browse/services/database/__init__.py
erickpeirson: > One question that remains though is whether, because the queries in each of these functions is executed once per request, whether we're introducing overhead for no benefit at all -- if a new baked query is getting constructed on each function call to `services.database`, then the cache isn't doing anything and clarity of code might be the bigger advantage.

Actually, that's not quite right (although it's certainly what one would expect given typical Python scoping rules).  From the docs:

> In the above code, even though our application may call upon search_for_user() many times, and even though within each invocation we build up an entirely new BakedQuery object, all of the lambdas are only called once. Each lambda is never called a second time for as long as this query is cached in the bakery.

> The caching is achieved by storing references to the lambda objects themselves in order to formulate a cache key;

In other words, there is a singleton behind the scenes somewhere that is storing a reference to the lambdas. Which...is pretty much what PEP8 is advising against.

erickpeirson: Here is an untested refactor of this function without the bakery that reduces the number of lines by ~18%.

```python
@db_handle_error(logger=logger, default_return_val=None) 
def get_sequential_id(paper_id: Identifier,
                      context: str = 'all',
                      is_next: bool = True) -> Optional[str]:
    """Get the next or previous paper ID in sequence."""
    local_session = db.session()
    if paper_id.is_old_id:
        # NB: classic did not support old identifiers in prevnext
        if context == 'all':
            like_id = f'{paper_id.archive}/{paper_id.yymm}%'
        else:
            like_id = f'%/{paper_id.yymm}%'
    else:
        like_id = f'{paper_id.yymm}.%'

    q = local_session.query(Document.paper_id) \
        .filter(Document.paper_id.like(like_id))

    if is_next:
        q = q.filter(Document.paper_id > paper_id) \
            .order_by(asc(Document.paper_id))
    else:
        q = q.filter(Document.paper_id < paper_id) \
            .order_by(desc(Document.paper_id))

    if context != 'all':
        archive: str = context
        subject_class: str = ''
        if '.' in context:
            archive, subject_class = archive.split('.', 1)
        q = q.join(in_category).filter(in_category.c.archive == archive)
        if subject_class:
            q = q.filter(in_category.c.subject_class == subject_class)

    result = q.first()
    if result:
        return f'{result.paper_id}'
    return None
```

mhl10: https://github.com/arXiv/arxiv-browse/pull/84

mhl10: Thanks for the summary of our discussion @erickpeirson. These are all good reasons to move away from baked queries, and refactoring is not a big deal. I'll resolve this conversation for now. Requested changes are reflected in #84.

erickpeirson:
@mhl10 @bdc34 Looking good; thanks for the changes

PR #82, opened 2019-04-08
Some changes to /archive 

bdc34:
Some changes to /archive that came up when comparing it to the legacy version.

Some bugs. Other changes were slight differences between NG and Legacy.


tests/legacy_comparison/archive_config.py
```diff
@@ -0,0 +1,12 @@
+from response_comparisons import compare_status
+
+res_comparisons = [ compare_status ]
+text_comparisons = []
+html_comparisons = []
+
+def ng_id_to_url_fn(id:str)->str:
+    return 'http://localhost:5000/archive/' + id
```
mhl10: Is this intentional?

tests/legacy_comparison/archives.txt
```diff
@@ -0,0 +1,20 @@
+astro-ph
```
mhl10: This will need to be updated independently of the taxonomy whenever there's a new archive.

tests/legacy_comparison/archive_config.py
bdc34: Yes. Is it a good idea to hard code that URL prefix? No, it should be a command line arg or something.

mhl10: For this function, it would probably be ok to use `url_for('archive', ...)`. But not a huge deal as it's in a test.

bdc34:
I talked with Martin and he said didn't think his comments on the legacy comparison scripts needed to be followed up on.

PR #83, opened 2019-04-12
Workaround for ARXIVNG-2063

mhl10:
This is just a temporary workaround for ARXIVNG-2063 and the case where a missing route in the URL map would cause an exception to get raised in the duplicate tapir_session cookie scenario.

### DIFF
--- a/browse/config.py
+++ b/browse/config.py
@@ -321,3 +321,9 @@ SESSION_DURATION = os.environ.get(
     'SESSION_DURATION',
     '36000'
 )
+
+URLS = [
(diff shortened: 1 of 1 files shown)


browse/config.py
```diff
@@ -321,3 +321,9 @@
     'SESSION_DURATION',
     '36000'
 )
+
+URLS = [
+    ('ui.login', '/login', os.environ.get('SERVER_NAME', 'arxiv.org'))
```
erickpeirson: This looks right to me

PR #92, opened 2019-04-29
ARXIVNG-2088: Browse should use the compactsearch macro

mhl10:
In addition to updating the browse base template to use the `compactsearch` macro from arxiv-base, this PR also adds support for `flask-s3`. Right now, the only file in the static folder is `browse_search.css` to support the NG search box; I'll move other static assets into this tree as part of a separate ticket.

@erickpeirson I think we'll need to expand on [this](https://github.com/arXiv/arxiv-base/blob/c99dda7943cca2d0f2803610b19beef97f608b7e/arxiv/base/__init__.py#L67) by amending the app's `url_map` for the static path (see this Stackoverflow [demo](https://stackoverflow.com/questions/26722279/how-to-set-static-url-path-in-flask-application)).

Screenshot 2019-04-29 13 35 53


browse/factory.py
```diff
@@ -13,10 +14,12 @@
 from arxiv.base import Base
 from arxiv.users.auth import Auth
 
+# s3 = FlaskS3()
+
 
 def create_web_app() -> Flask:
     """Initialize an instance of the browse web application."""
-    app = Flask('browse', static_folder='static', template_folder='templates')
+    app = Flask('browse', static_url_path='/static')
```
bdc34: What happened to the template parameter?

mhl10: `templates` is the default value, so that change shouldn't affect anything. I'm still messing a bit with flask-s3, so am not quite ready yet.

bdc34: The legacy system has a problem where the static assets are difficult to develop with. 
It would be great only the production system served assets from S3.

mhl10:
> I have minor concerns about browsers caching /abs pages, getting a 304 Not Modified since the abs was not modified, then not being able to get /static/browse/0.2.2/css/browse_search.css because we are now on /static/browse/0.3.1/css/browse_search.css

Good catch! I hadn't thought of that. Right now, we're preserving the static files for previous versions. There may be some oddities for people revisiting the same abs page within ~24 hours, but I expect they'll be pretty minor.

PR #93, opened 2019-05-01
ARXIVNG-2117, ARXIVNG-2118: Remove CiteULike, "what is this?" links from /abs bookmarks

mhl10:


mhl10:
> Looks OK. Could possibly drop the "what-is-this" class in the div, but I'm not certain. Fine as is.

Thanks for catching this. I'll remove this in another branch.

PR #98, opened 2019-06-10
ARXIVNG-1728: reimplementation of category taxonomy page

mhl10:
Per discussion with @jimentwood, we decided to reimplement the classic taxonomy in browse in the short term, as /category_taxonomy.  At some point we can move this to arxiv-docs, under /about; some additional work will be needed here to support the dynamic nature of this page.

The only difference between this reimplementation and the classic page /mod/category_taxonomy is that the group no longer contains the group ID; the way the classic page was doing this wasn't really correct.

Screenshot 2019-06-10 15 44 54


browse/controllers/stats_page/__init__.py
```diff
@@ -121,6 +121,8 @@ def get_monthly_submissions_page() -> Response:
     try:
         num_submissions, historical_delta = \
             get_monthly_submission_count()
+        num_this_month = get_document_count_by_yymm(current_dt.date)
```
bdc34: Are these changes intended to be part of this PR or are the part of some other work?

mhl10: Not originally. This a "miscellaneous fix" for a bug that I just happened to notice while I was working on this other stuff.

PR #102, opened 2019-06-18
ARXIVNG-1985: proxy field in Submission history should (mostly) show full text

mhl10:
Some background first: 

- as of today, there are 11,270 distinct proxy names in the system
- if you remove the proxy names that start with "ccsd", you are left with 130 distinct names
- of the 11,140 proxy names that start with "ccsd", the names generally look like this (where `N` is an integer):
  `ccsd ujm-NNNNNNNN`
  `ccsd hal-NNNNNNNN`
  `ccsd inria-NNNNNNNN`

The classic display logic simply took the first word in the proxy name and capitalized it; I think this makes sense for the CCSD case but not necessarily others. So, in this PR I'm proposing the following changes to the display logic:
- if the proxy name is only one word long OR starts with "ccsd", then upper case it
- if the proxy name is more than one word long, display it as is, with tex2utf filtering; trail with "as proxy" instead of just "proxy"

Examples "proxy name" -> display:
- "ccsd hal-00460810" -> [via CCSD proxy]
- "ccsd inria-00460912" -> [via CCSD proxy]
- "vtex" -> [via VTEX proxy]
- "Martin Wolf" -> [via Martin Wolf as proxy]
- "Hajnal Andr\'eka" -> [via Hajnal Andréka as proxy]

### DIFF
--- a/browse/templates/user_banner.html
+++ b/browse/templates/user_banner.html
@@ -1,3 +1,5 @@
+{# donation banner #}
+{%- if 0 -%}
 <div class="slider-wrapper" style="display:none">
   <a class="close-slider" href="#"><img src="{{ url_for('static', filename='images/icons/close-slider.png') }}"></a>
   <div class="copy-donation">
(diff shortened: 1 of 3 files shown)


browse/templates/user_banner.html
```diff
@@ -1,3 +1,5 @@
+{# donation banner #}
+{%- if 0 -%}
```
bdc34: Wasn't this part of an earlier PR too? Or maybe this isn't important because that is already in develop?

mhl10: I don't think so?

mhl10: Oh I see, my bad!

PR #103, opened 2019-06-19
Pre-release merge for v0.2.4

mhl10:
ARXIVNG-1985 - show full proxy name under certain circumstances

Also includes news/banner updates and issue template markdown for github/jira integration.

### DIFF
--- a/browse/templates/abs/abs.html
+++ b/browse/templates/abs/abs.html
@@ -53,7 +53,7 @@
       <h2>Submission history</h2> From: {{ abs_meta.submitter.name|tex2utf if abs_meta.submitter.name != None }} [<a href="/show-email/{{ abs_meta.arxiv_id|show_email_hash }}/{{ abs_meta.arxiv_id }}">view email</a>]
       {#- Extra message for proxy sites (i.e. Proxy line has username and id) -#}
       {#- TODO: revisit this. Logic for display follows classic but is strange; sometimes a proxy is just a person's name. -#}
-      {% if abs_meta.proxy != None and abs_meta.proxy|wordcount > 1 %}&nbsp;[via {{ abs_meta.proxy.split()[0]|tex2utf|upper }} proxy]{% endif %}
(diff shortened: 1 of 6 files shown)


browse/templates/abs/abs.html
```diff
@@ -53,7 +53,7 @@ <h1>{% if abs_meta.primary_archive.id != abs_meta.primary_category.id %}{{ abs_m
       <h2>Submission history</h2> From: {{ abs_meta.submitter.name|tex2utf if abs_meta.submitter.name != None }} [<a href="/show-email/{{ abs_meta.arxiv_id|show_email_hash }}/{{ abs_meta.arxiv_id }}">view email</a>]
       {#- Extra message for proxy sites (i.e. Proxy line has username and id) -#}
       {#- TODO: revisit this. Logic for display follows classic but is strange; sometimes a proxy is just a person's name. -#}
-      {% if abs_meta.proxy != None and abs_meta.proxy|wordcount > 1 %}&nbsp;[via {{ abs_meta.proxy.split()[0]|tex2utf|upper }} proxy]{% endif %}
+      {% if abs_meta.proxy != None %}&nbsp;[via {% if abs_meta.proxy|wordcount == 1 or abs_meta.proxy.startswith('ccsd') %}{{ abs_meta.proxy.split()[0]|upper }} proxy{% else %}{{ abs_meta.proxy|tex2utf }} as proxy{% endif %}]{% endif %}
```
erickpeirson: Really hate to see ``ccsd`` hard-coded here. What is the context for this, and is there a way we can make this configurable?

mhl10: Some additional context for the `ccsd` case is in #102. I agree hard-coding is suboptimal; right now, this is the only case like this and I don't expect others would be added.

erickpeirson: Gotcha. So two things that would be worth doing:

1. Add an entry to DECISIONS.md with the description from #102 
2. Add a comment in ``abs.html`` that references the entry in DECISIONS.md

That way at least we have a clear record of what's going

erickpeirson: (I see that there isn't a DECISIONS.md, so I guess step 0 is to create one ;-) )

mhl10: Sounds good.

mhl10: @erickpeirson DECISIONS.md now added.

erickpeirson: Awesome, thanks!

PR #121, opened 2019-10-31
ARXIVNG-2758 various accessible edits

SBBCornell:
Addresses the following accessibility related edits on browse:
- fixed homepage links by adding unique aria labels.
- addressed duplicate IDs on abs.
- added secondary search form aria labelling to homepage.

Please note the changes to the abs page are dependent on a very small PR for base (single line edit made to templates/base/macros.html)


SBBCornell:
these changes should not be deployed without the latest edit to base (a one line change in templates/base/macros.html)

PR #125, opened 2019-11-19
ARXIVNG-2779 mobile friendly abs page

SBBCornell:
mobile friendly abs page edits are ready for deployment on dev or beta, whichever is best for wider user testing.

### DIFF
--- a/browse/templates/abs/extra_services.html
+++ b/browse/templates/abs/extra_services.html
@@ -14,6 +14,7 @@
   {% endif %}
 {%- endmacro -%}
 
+<a name="other-sources"></a>
 {%- macro generate_download_links(format_list=[]) -%}
 <ul>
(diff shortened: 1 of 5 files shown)


browse/templates/abs/extra_services.html
```diff
@@ -161,6 +162,7 @@ <h3><a href="{{ dblp.base_url }}">DBLP</a> - CS Bibliography</h3>
     </script>
 {% endif %}
     <div class="full-text">
+      <a name="other"></a>
```
bdc34: Is this a mistake?

SBBCornell: No, no a mistake, but thanks for double checking. Thats just an anchor link that the mobile button jumps the user to if there is no PDF.

PR #126, opened 2019-11-19
ARXIVNG-2794 accessible browse footer icons

SBBCornell:
make browse footer icons accessible, switched to using svg icons, self hosting the font-awesome package, and made some minor edits to more closely match the footer in _base.


browse/static/fontawesome-svgs/README.txt
```diff
@@ -0,0 +1,19 @@
+How to use this implementation of Font Awesome
```
mhl10: Could this be converted to markdown, and renamed to README.md?

PR #137, opened 2020-01-24
updating to python 3.8; fixing a few lowhanging type issues

bbarker:
- [x] Ran tests
- [x] Confirmed page could load when running on local host with test db
- [x] All tests seemed to pass (22 were skipped for some reason, guessing related to db availability)
- [ ] all mypy tests pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
 os:
   - linux
 python:
-  - "3.6"
+  - "3.8"
 script:
   - pip install pipenv
   - pipenv sync --dev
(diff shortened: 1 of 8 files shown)


bbarker:
looks like I need to get travis to use 3.8, will check on that shortly

mhl10:
Hi @bbarker, did you know whether 3.8 implies the latest point release (which I think is 3.8.1? 
As an aside, we'll need to install a new reference copy of python and probably rebuild mod_wsgi if we want to test this on production. We could test this on beta.

mhl10:
closing this for now as there's been no follow-up

PR #138, opened 2020-02-03
ARXIVNG-2902: move Google and Semantic Scholar links into dedicated sidebar section

mhl10:
The screenshot below shows the sidebar with a new "Additional Resources" section that contains outbound search links for Google Scholar and Semantic Scholar. Previously, the Google Scholar link was being injected by the bibex javascript in the same section with the "Export citation" link. 

My questions:
- Is "Additional Resources" the best name for this new section?
- Perhaps not germane to this PR, but what if anything should we do about the "Export citation" link that is still being injected by bibex (regardless if bibex is enabled or not). It does seem valuable, but should probably be grouped with the other bibex elements if we want a clearer separation between core arXiv functionality and arXiv Labs functionality.

Changes are staged on beta, e.g. https://beta.arxiv.org/abs/1911.11778

Screenshot 2020-02-03 14 53 30

Edit: moved links into "References & Citations" section:
Screenshot 2020-02-03 15 55 34

### DIFF
--- a/browse/templates/abs/extra_services.html
+++ b/browse/templates/abs/extra_services.html
@@ -213,6 +213,17 @@
       {{ generate_dblp_section() }}
     {%- endif -%}
 
+    <div id="additional-resources" class="extra-ref-cite">
+      <div><h3>Additional Resources</h3></div>
+        <ul>
+        <li>
+          {#- This was previously injected by Bibliographic Explorer -#}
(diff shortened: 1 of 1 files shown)


jweiskoff:
How about something like "View metadata on..." since it's more clear that's what they're being redirected to?

mhl10:
> You could then also group the ADS link there, but maybe that's not what we're after...?

I think that could work.

mhl10:
> > You could then also group the ADS link there, but maybe that's not what we're after...?
> 
> I think that could work.

I've updated the branch (and beta) with the changes; links are now in "References & Citations".

jimentwood:
I like it

PR #147, opened 2020-03-13
ARXIVNG-3025 fixes last test errors

SBBCornell:
minor edits to fix testing errors on the trackback pages. Also includes minor style edits.

### DIFF
--- a/browse/templates/tb/tb.html
+++ b/browse/templates/tb/tb.html
@@ -27,7 +27,7 @@
     {% for tb in trackback_pings %}
   <p>{{- generate_trackback_link(tb) -}}</p>
     {% endfor %}
-    <blockquote style="background-color: #eee; padding-top: 5px">
+    <blockquote>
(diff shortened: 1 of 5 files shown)


browse/templates/tb/tb.html
```diff
@@ -38,5 +51,5 @@ <h2>Trackbacks for <a href="{{ url_for('.abstract', arxiv_id=arxiv_identifier.id
   {% else %}
   <p><em>There are no trackback pings recorded for <a href="{{ url_for('.abstract', arxiv_id=arxiv_identifier.id) }}">{{ arxiv_identifier.id }}</a>.</em></p>
   {% endif %}
-  {{ super() }}
```
bdc34: The TB page looks good.
I just wanted to check that this is an intentional removal of this call to the parent template.

SBBCornell: it is intentional. That call was just moved to a different containing block for layout reasons.

PR #170, opened 2020-07-06
trackbacks: emphasis on title

SBBCornell:
Some html and css edits to emphasize the trackback title more and the metadata less.

When spinning it up locally I cant find a paper that has a trackback. The attached are tested with these changes in-browser on the live site, but Id love to locate a test paper with a trackback to test with also before it is pushed live. Do you know of any?

Screenshot of the new visual changes for reference:
Screen Shot 2020-07-06 at 4 32 50 PM

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "2a97ec36f67169ec6b8c818e2a85c427316579a3d8bdb233e474607800b2ee1e"
+            "sha256": "fdc7d0cbc02d677bf2efe5597e911e2064b83f406a1278ee236c4b8ab5c38663"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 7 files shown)


mhl10:
@SBBCornell [0808.4142](http://localhost:5000/abs/0808.4142) is the example with [trackbacks](http://localhost:5000/tb/0808.4142). You may need to run `FLASK_APP=app.py pipenv run python populate_test_database.py --drop_and_create` in case you don't see it.

Screenshot 2020-07-06 21 26 13

The bigger trackback title looks good. I wonder about the paper metadata still, which was a carryover from the old trackback page. Users who are viewing the trackback page for an individual paper are usually clicking from the /abs page for that paper, so it may not be as useful to include the paper metadata so predominantly. Maybe the metadata could be shown when hovering over the paper ID in "Trackbacks for 0808.4142". Or an expand/collapse widget.

I would also lean towards enlarging "Trackbacks for 0808.412" with `h1` (it's `h2`now).

Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
```
mhl10, line 1: This file is probably ok as committed, but probably not needed for these changes

SBBCornell:
Updated screenshot with changes from latest commit. The metadata is now hidden by default and can be toggled open:
Screen Shot 2020-07-07 at 9 29 58 AM

SBBCornell:
latest commit reduces emphasis on the title in the trackback guide block:

Screen Shot 2020-07-10 at 10 07 48 AM

mhl10:
@SBBCornell, while I was fixing tests I made a few other minor changes:
- fix the vertical gap on the /abs page caused by the what-is-this style
Screenshot 2020-07-22 10 51 43

- add the trackback-styles class to the main div in the /tb/recent page:

Screenshot 2020-07-22 10 50 57

PR #171, opened 2020-07-10
sets max width of CORE div to 100%

SBBCornell:
very minor edit. Sets max width of #coreRecommenderOutput to 100%.


SBBCornell:
branch not needed, added to ARXIVNG-3176

PR #172, opened 2020-07-15
Clean-up CORE Recommender script

viktor-yakubiv:
1. Removes cookie management from the `core-recommender.js`
2. Adds a conditional check for the script loading:
   - if the `document` was loaded already, the `init()` function
     is executed
   - if the `document` was not loaded, a new 'load' listener is added

---

I checked how the arXiv Labs' switchers work and found that the cookie management is done separately. I removed this from the CORE Recommender's stript accordingly.

Also, I fixed the problem with immediate loading of the recommendations. There was a little unexpected case.

However, I see the differences between the deployed script on the `beta` and the latest version in the `develop` branch. I hope there will be no conflicts.

---

I will properly test it and turn off the draft later.

### DIFF
--- a/browse/static/js/core-recommender.js
+++ b/browse/static/js/core-recommender.js
@@ -6,39 +6,8 @@
 
   var document = window.document;
   var baseUrl = 'https://core.ac.uk';
-  var cookieName = options.cookieName || "arxiv_core_recommender";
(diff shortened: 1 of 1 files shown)


mhl10:
Looks good, we'll be sure to use this version. Thanks @viktor-yakubiv !

PR #180, opened 2020-08-10
ARXIVNG-3176: MVP for new Labs design and toggle functionality

mhl10:
We'd like like to get the new Labs design and toggle functionality deployed before we add new Labs features (like CORE Recommender):

Example: https://dev.arxiv.org/abs/2006.13744

Screen Shot 2020-08-10 at 3 39 57 PM


mhl10:
@SBBCornell I just noticed that the browse feedback link ("Browse v0.3.2.5 released ...") bleeds into the Labs section when Labs is enabled. Could that be fixed?

SBBCornell:
we blazed past this with other branches!

PR #181, opened 2020-08-19
ARXIVNG-3458: labs tabs higher in abs

mhl10:
This PR adds markup for tabbed display higher in the abs page, where the abstract field is moved into a tab and set to the default. This style can be toggled with the `labs-style` query parameter set to `tabs-above`:

http://127.0.0.1:5000/abs/0704.0053?labs-style=tabs-above
Screen Shot 2020-08-18 at 9 29 08 PM

This `tabs-above` implementation uses several hacks to reuse the `abs` macro from arxiv-base; it is not suitable for production but should be suitable for testing purposes. 

The `tabs-below` style can be toggled:
http://127.0.0.1:5000/abs/0704.0053?labs-style=tabs-below
Screen Shot 2020-08-18 at 9 28 33 PM

Finally, the original page without any Labs elements can be toggled without any parameters:
http://127.0.0.1:5000/abs/0704.0053

### DIFF
--- a/browse/static/js/toggle-labs.js
+++ b/browse/static/js/toggle-labs.js
@@ -1,6 +1,11 @@
 // vanilla toggle script for LABS enabling
 $(document).ready(function() {
 
+  // TODO: Remove this hack if macro for tabs-above style is created in arxiv-base
+  if ( $('.dateline').size() > 1 ) {
+    $('.dateline').last().remove();
+  }
+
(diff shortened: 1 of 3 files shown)


mhl10:
> Could the mobile download button be moved above the tabs? On the live site the download button is above the abstract description.

@SBBCornell change pushed!

PR #186, opened 2020-09-07
PaperWithCode lab widget

ludoviofb:
This is the implementation of the lab widget for PaperWithCode.
This is a first draft for testing purpose and will be enough for us to test our integration.
For production we would need a way to only activate the lab on papers related to machine learning.


mhl10 (approved):
These changes look good, thanks @ludoviofb. Once merged we can stage changes on one of our development servers.

PR #192, opened 2020-09-24
News item about TeX updates on homepage

SBBCornell:
This branch adds a news item to the homepage about the TeX live update.

This is the message that will display up until 7:55am on Oct. 1:
Screen Shot 2020-09-24 at 4 31 54 PM

After that date/time passes and up until the end of Nov. the news section will display this message:
Screen Shot 2020-09-24 at 4 43 55 PM

(informational aside: the original plan was to add this blurb with Pendo only but I learned that they have very limited options for inline text injection. Their guides are intended to float on top of existing content as a new layer, or be a full width badge at the top or bottom of the page. I can inject a 'badge' but it requires special css tweaks that target the unique pendo badge ID, which means pushing an edit to our servers anyway. It also launches a guide when clicked, though I can make that guide invisible, but it still feels hacky. I will experiment with that further but for now we can add this news via this branch instead.)

### DIFF
--- a/browse/static/css/arXiv.css
+++ b/browse/static/css/arXiv.css
@@ -2488,12 +2488,12 @@ body#front.with-cu-identity #header #search {
 
 
 /**** Pendo Guide Badge Styles - when you only want a badge/text note and no corresponding guide ****/
-.intro-and-news .bb-text {
+/* .intro-and-news .bb-text {
     text-align: left !important;
 }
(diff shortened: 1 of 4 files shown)


ep538 (approved):
looks good!

afromme (approved):
Looks good!

SBBCornell:
The news section is updated with a blog link:
Screen Shot 2020-09-25 at 2 02 36 PM

PR #194, opened 2020-10-14
Papers with code widget

ludoviofb:
Add the possibility to have multiple official code implementations for papers.
Group the css to the global file for abstract to avoid having to make an other query and having async css just for 20 lines


browse/static/css/arXiv.css
```diff
@@ -827,6 +827,31 @@ p.tagline {
   cursor: pointer;
 }
```
SBBCornell, line 829: add a note for future editors:
```suggestion
/*added by Papers With Code*/
```

browse/static/css/arXiv.css
```diff
@@ -827,6 +827,31 @@ p.tagline {
...
+}
+
+.pwc-icon-primary {
+    top: 6px;
+}
+
+.pwc-avatar-box {
+    display: inline-block;
+}
+
```
SBBCornell, line 854:
```suggestion
/*END Papers With Code css*/
```

mhl10 (approved):

PR #198, opened 2020-11-23
Arxivng-3649 pendo add logged in user

bmaltzan:
This will write the institution_id and encoded user_id into the pendo snippet, as the account and visitor. It adds geo data from ip address. It's on dev now.

To test visitor.id on dev, add your ip to the aws-security group and devs vhost.conf. And update your local /etc/hosts to point to dev's public ip.
To test account.id locally, there's a localhost-university


browse/templates/base.html
```diff
@@ -37,7 +57,12 @@
               },
 
               account: {
-                  id:           'ACCOUNT-UNIQUE-ID' // Highly recommended
+                  {% if session.get("institution") %}
+                  name:            '{{ session.get("institution") }}',
+                  {% endif %}
+
+                  id:              '{{ session.get("institution_id") if session.get("institution_id") else "ACCOUNT-UNIQUE-ID" }}'
```
bdc34, line 64: Do we really need to put ACCOUNT-UNIQUE-ID here or is that just documentation from Pendo to indicate the kind of thing to put as the value here?

tests/test_database_service.py
```diff
@@ -92,16 +92,16 @@ def setUpClass(cls) -> None:
     def test_get_institution_returns_a_label(self) -> None:
         """If IP address matches an institution, a label is returned."""
         label = TestBrowseDatabaseService.database_service.get_institution(
-            '128.84.0.0')
+            '128.84.0.0').get('label')
```
bdc34, line 95: These could test ID too or is that never used?

bdc34 (approved):
I have a couple very minor questions but nothing that should prevent this from being merged and deployed.

SBBCornell (approved):
it looks great, congrats on getting to this point!

browse/controllers/archive_page/__init__.py
```diff
@@ -72,15 +72,17 @@ def archive_index(archive_id: str, status: int) -> Tuple[Dict[str, Any], int, Di
         for id in ARCHIVES.keys()
         if id not in ARCHIVES_SUBSUMED and not id.startswith("test")
     ]
-    archives.sort(key=lambda tpl: tpl[0])
+    k = lambda tpl: tpl[0]
```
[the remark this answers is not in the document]
bmaltzan: Nose logs an error for that syntax. I think it was a change in mypy==0.790

browse/templates/base.html
bmaltzan: It has been deployed that way in other places, and shows up in pendo's ui. Seems fine to remove or keep.

tests/test_database_service.py
bmaltzan: Id is used, so I'll add a test.

browse/templates/base.html
bmaltzan: It looks like a value for visitor_id is required, but account_id is not. I think there's value in having Pendo account pages show how many visitors are not associated with an account/institution, so I think we should keep some constant here. We could add another localhost type of value to the subscription db tables or set it as -1. That seems cosmetic, so I think we leave it with their default.

browse/controllers/archive_page/__init__.py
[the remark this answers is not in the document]
bmaltzan: I've reverted and told mypy to ignore these lines. That seems preferable vs disabling the check.

PR #204, opened 2020-12-18
This fixes etag quotes and a couple other things

bdc34:
This fixes etag handling and last-modified time handling.

It also gets the request headers for these in a case insensitive way.

Jim and I were hoping that this would fix people needing to refresh to get updated papers. The refresh would be intended to get around the sever returning a 304 Not-Modified. But this bug was causing the server to always 200 (Ok) and never 304 (Not modified).

### DIFF
--- a/browse/controllers/abs_page/__init__.py
+++ b/browse/controllers/abs_page/__init__.py
@@ -196,7 +196,7 @@ def _check_request_headers(docmeta: DocMetadata,
     
     not_modified = _not_modified(last_mod_dt,
                                  _time_header_parse('If-Modified-Since'),
-                                 request.headers.get('If-None-March', None),
+                                 _get_req_header('if-none-match'),
                                  etag)
 
     return not_modified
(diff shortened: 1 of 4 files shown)


browse/controllers/abs_page/__init__.py
```diff
@@ -82,6 +82,11 @@ def get_abs_page(arxiv_id: str) -> Response:
             return redirect
 
         abs_meta = metadata.get_abs(arxiv_id)
+        not_modified = _check_request_headers(
```
bdc34, line 85: Moved the not-modified check earlier to avoid unnecessary computation.

browse/controllers/abs_page/__init__.py
```diff
@@ -189,50 +187,51 @@ def _check_request_headers(docmeta: DocMetadata,
...
-    if mod_since_dt and none_match_dt:
-        not_modified = (mod_since_dt >= last_mod_dt
-                        and none_match_dt >= last_mod_dt)
-    elif mod_since_dt and not none_match_dt:
-        not_modified = mod_since_dt >= last_mod_dt
-    elif none_match_dt and not mod_since_dt:
-        not_modified = none_match_dt >= last_mod_dt
+                  none_match: Optional[str],
+                  current_etag: str) -> bool:
+    if none_match and none_match == current_etag:
```
bdc34, line 209: I changed this logic a bunch and put the etag first because it is suppose to take priority over last-modified.

browse/controllers/abs_page/__init__.py
```diff
@@ -189,50 +187,51 @@ def _check_request_headers(docmeta: DocMetadata,
...
+    except (ValueError, TypeError, KeyError):
+        pass
+    return None
+
+
+def _get_req_header(header: str) -> Optional[str]:
+    """Gets request header, needs to be case insensative for keys.
+
+    HTTP header keys are case insensative. RFC 2616"""
+    return next((value for key, value in request.headers.items()
```
bdc34, line 233: The HTTP header keys need to be accessed in a case insensative way. So if the client can send If-None-Match or if-none-match or If-NoNe-mAtCh.

browse/controllers/abs_page/__init__.py
```diff
@@ -189,50 +187,51 @@ def _check_request_headers(docmeta: DocMetadata,
...
-    elif mod_since_dt and not none_match_dt:
-        not_modified = mod_since_dt >= last_mod_dt
-    elif none_match_dt and not mod_since_dt:
-        not_modified = none_match_dt >= last_mod_dt
+                  none_match: Optional[str],
+                  current_etag: str) -> bool:
+    if none_match and none_match == current_etag:
+        return True
+    elif mod_since_dt:
+        return (mod_since_dt.replace(microsecond=0) >=
```
bdc34, line 212: The last_mod_dt had microseconds and  mod_since_dt was always zero microseconds. 

So when there was no change to the paper, the mod_since would be less than last_mod and the server would do a 200 when it should have done a 304.

jimentwood (approved):
I can't really judge the code changes, but the concept makes sense.

DavidLFielding (approved):
Looks good.

browse/services/util/response_headers.py
```diff
@@ -12,12 +12,17 @@
 """Days of the week publish happens: Sunday-Thursday."""
 
 
-def guess_next_update_utc(dt: datetime = datetime.now(timezone.utc)) \
-        -> Tuple[datetime, bool]:
+def guess_next_update_utc(dt: Optional[datetime] = None) -> Tuple[datetime, bool]:
```
bdc34, line 15: Changed default parameter to none so it gets set to now.

When set in the function definition, dt get set to the time the function definition is evaluated.

bdc34, line 15: Changed default parameter to none so it gets set to now.

When set in the function definition, dt get set to the time the function definition is evaluated.

PR #209, opened 2021-02-01
Some edits to labs-in-tabs

SBBCornell:
This branch includes a few changes to labs-in-tabs-on-abs-so-fabs:

The About tab is moved to the last position and the content is updated to two columns, with the labs icon in the far column. Its responsive and stacks on mobile. The tab "Bibliographic Tools" is now first and the default selected tab.
Screen Shot 2021-02-01 at 2 43 14 PM

There are also some changes to the layout under 'Related Papers' (we've combined both toggles into the same blue area), and some accessibility fixes.
Screen Shot 2021-02-01 at 2 43 24 PM


mhl10 (approved):
Looks great!

PR #210, opened 2021-02-02
Connected papers - fix minor css bugs

KanHarI:
Hi,

During testing we detected a minor bug in our hover effect on some displays and also a misalignment between our css and yours for the threshold of switching between desktop and mobile versions, fixed both


mhl10 (approved):
Good catch -- thanks!

PR #216, opened 2021-03-23
Bad abs 2103 11058 hotfix

bdc34:
See https://arxiv-org.atlassian.net/browse/ARXIVNG-3976

The double backslashes were causing problems. They are used in the abs file for both section dividers and LaTeX line breaks. 

I switched from a regex to split these to looping over the lines. This will tolerate of when a submitter ends up getting a double backslash on a line all by its self.


DavidLFielding (approved):
This looks good to me. 

The first '\\' metadata/abs delimiter is allowed to fall through to the abs, but then you clean it up later.

Is this deployed on a development server with problematic abstract?

mhl10 (approved):

PR #218, opened 2021-04-28
Combined Code & Data tab via Papers with Code

rstojnic:
This is a version of #217 where the both code and data are shown in a single tab, with a single component. Also it's using a new API endpoint to fetch both code and data in a single API request. The Datasets section is always hidden for non-ml/cs/stat.


bmaltzan (approved):

mhl10 (approved):
Looks great!

Is the value N in "N papers also use this dataset" referring to arXiv papers or your entire corpus? If the latter, it would be nice to also break this out with the number of papers in arXiv. Bonus to also include a direct link to a list of those N papers at some point :-) 

I also noticed that the untoggled/disabled state isn't persisting with reloads. It looks like this bug predates this PR, so we'll fix that after this PR is merged.

rstojnic:
@mhl10 it refers to our whole corpus, we added a list of papers to our dataset page - just hasn't been deployed yet. We could separate our the number of arXiv papers, but I think this would make more sense once you can do something with this information (i.e. click on it to lead somewhere on arXiv?).

I'll have a look at saving the toggling state - I remember it used to work.

rstojnic:
@mhl10 have a look at 41c92acce23f4cc4ff287e9377d2f12802eea284 - I think I fixed the issue with the widget not switching off when when it's switched off in the cookie.

mhl10:
> @mhl10 have a look at [41c92ac](https://github.com/arXiv/arxiv-browse/commit/41c92acce23f4cc4ff287e9377d2f12802eea284) - I think I fixed the issue with the widget not switching off when when it's switched off in the cookie.

Thanks @rstojnic, I'll merge this and get it staged on one of our dev servers.

PR #224, opened 2021-05-26
Add “Open in Litmaps” integration to labs

axtonpitt:
Changes made to add "Open in Litmaps" functionality to abstract pages. I have tried to mirror other integrations to get the integration right.

Signed-off-by: Axton Pitt <axton@litmaps.co>

### DIFF
--- a/browse/static/js/litmaps.js
+++ b/browse/static/js/litmaps.js
@@ -1,5 +1,5 @@
 (function () {
-    const arxivId = window.location.pathname.split('/').reverse()[0];
+    const arxivId = document.head.querySelector("[name~=citation_arxiv_id][content]").content;
     const litmapsBaseUrl = 'https://api.litmaps.co';
     const litmapsAppUrl = 'https://app.litmaps.co';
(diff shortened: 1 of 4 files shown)


browse/static/js/litmaps.js
```diff
@@ -0,0 +1,47 @@
+(function () {
+    const arxivId = window.location.pathname.split('/').reverse()[0];
```
mhl10, line 2: Hi @axtonpitt, I noticed when accessing a specific version of a paper (i.e. when the arXiv ID in the path has a version affix), Litmaps is less successful with the lookup. So I would exclude the version affix from all lookups. Example:

https://beta.arxiv.org/abs/2001.00001v2

You can get the canonical ID from one of the metatags as follows:
```suggestion
    const arxivId = document.head.querySelector("[name~=citation_arxiv_id][content]").content;
```

mhl10:
Looks great @axtonpitt. I had one minor suggestion above.

axtonpitt:
Ok thanks @mhl10, have committed your suggested change.

mhl10 (approved):

PR #231, opened 2021-07-14
ARXIVNG-4282: Implement version handling in DB backend abs

DavidLFielding:
This ticket implements version handling on the DB backend side. In the legacy system the .abs file contains all of the version information we need. On the DB side the version information is stored with each version record.

This work is basically complete EXCEPT for the timezone specification in the submission history. There is NO timezone provided with the File System implementation of version yet the db test (that fails) is looking for one. The production site shows the timezone.

[v1] Thu, 11 Jun 2009 14:09:14 UTC (20 KB)

I want to verify that this wasn't removed intentionally before I try to add it back. Submission date from DB gets nicely formatted somewhere so I'm wondering if we are missing an option where this happens.


mhl10 (approved):
Looks good!

PR #235, opened 2021-07-27
ARXIVNG-4258: dissemination formats from DB metadata

mhl10:


browse/factory.py
```diff
@@ -6,7 +6,7 @@
 from arxiv.base.urls import canonical_url, clickthrough_url, urlizer
 from arxiv.base.config import BASE_SERVER
 from arxiv.base import Base
-from arxiv.users.auth import Auth
+# from arxiv.users.auth import Auth
```
bdc34, line 9: This commented out due to the "Restore support of login status in Browse" issue.
https://arxiv-org.atlassian.net/browse/ARXIVNG-4354

browse/services/documents/db_implementation/db_abs.py
```diff
@@ -133,25 +123,28 @@ def get_dissemination_formats(self,
...
+                else:
+                    dissemination_formats.append('sciencewise_pdf')
+
+        else:
+            return formats_from_source_type(format_code,
+                                            format_pref,
+                                            False,
+                                            add_sciencewise)
+
         return []
```
bdc34, line 148: Is this return[] extraneous?

bdc34 (approved):
Seems good.

browse/services/documents/db_implementation/db_abs.py
mhl10: It is!

PR #240, opened 2021-08-25
ensure scite badge uses right aligned tooltip placement

domenicrosati:
# Change

Use the tooltip placement attribute to ensure the scite badge is right-aligned. In this way the stacking and overlap issues are mitigated.

Screen Shot 2021-08-25 at 1 10 28 PM

cc 👁️  @mhl10


mhl10 (approved):
Thanks for the speedy turnaround!

PR #255, opened 2022-02-02
Fix Docker instructions in README

zeke:
This PR updates the Docker section of the README.

Before this change, running the second command fails:

```
$ docker run -it 8000:8000 arxiv/browse:some_tag          
Unable to find image '8000:8000' locally
docker: Error response from daemon: pull access denied for 8000, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
```

Summary of changes:

- Updates the second run command to include the missing `--publish` flag
- Removes the `:some_tag` suffix from the image name. It's not strictly needed and could cause confusion.
- Adds a link to the Docker download page
- Removes a code block syntax error at the bottom of the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,3 +113,4 @@ Goal: 9/10 or better.
 
 ```bash
 pipenv run pylint browse
+```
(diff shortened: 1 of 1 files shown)


zeke:
cc @SBBCornell and @mhl10, my arXiv Labs liaisons :)

mhl10 (approved):
Thanks for catching this!

PR #259, opened 2022-02-14
Logo changes

mhl10:


SBBCornell (approved):
I see there are some test files. I didnt review those because I dont know their purpose. I only reviewed the html and css files related to the logo change and all looks as expected.

PR #257, opened 2022-02-02
Display machine learning demos from Replicate.com

zeke:
This pull request introduces a new [arXiv Labs](https://labs.arxiv.org/) integration that displays links to interactive machine learning demos from @replicate.

## Implementation notes

- I used the existing PapersWithCode integration as a reference implementation.
- There's a new "Demos" tab has between "Code & Data" and "Related Papers". (I believe this was a suggestion from the arXiv team during early design discussions?)
- Favors well-supported browser APIs (doesn't use any jQuery)
- The generated HTML is sanitized to prevent XSS

## Screenshots

Demo exists:

Screen Shot 2022-02-03 at 9 45 02 PM

---

No demo exists:
Screen Shot 2022-02-03 at 9 45 45 PM

---

Integration disabled:
Screen Shot 2022-02-03 at 9 45 57 PM

## Running the development enviroment

```
docker build . -t arxiv/browse
docker run -it --publish 8000:8000 arxiv/browse
```

## Viewing the integration in development

The arxiv/browse dev environment includes a small subset of articles, but it appears that they're all pretty old, and none of the papers for which Replicate models exist are present in the dev environment. To work around this, I made it possible to inject a custom query param onto the URL to override the arxiv paper ID.

To see how demos are rendered, try these example URLs:

- http://localhost:8000/abs/0704.0182?override_paper_id=2103.17249
- http://localhost:8000/abs/0704.0182?override_paper_id=2101.04061
- http://localhost:8000/abs/0704.0182?override_paper_id=2108.10257

To see what the "no models exist" state looks like, visit any page (without setting the query param):

- http://localhost:8000/abs/0704.0182

PR #261, opened 2022-02-15
Add start script

zeke:
In my recent work on #257, I noticed that I had to stop the server and run these Docker commands every time I made changes. This PR adds a start script that bundles the commands into one, which makes it a bit easier to start and stop the server.

The script is following the [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all) methodology popularized by GitHub.

cc @mhl10 @SBBCornell


mhl10 (approved):
Looks good!

PR #264, opened 2022-02-18
Add a meta tag exposing an article's DOI

Vinnl:
This will make it discoverable by Google Scholar and other
indexers. See e.g.
https://scholar.google.com/intl/it/scholar/inclusion.html#indexing


mhl10 (approved):
Thank you for your contribution!

PR #268, opened 2022-04-12
ARXIVNG-4916-iarxiv

bmaltzan:
Here's a branch for the IArxiv labs updates.

### DIFF
--- a/browse/templates/abs/labs_tabs.html
+++ b/browse/templates/abs/labs_tabs.html
@@ -164,13 +164,18 @@
         <div class="columns is-mobile lab-row">
           <div class="column lab-switch">
             <label class="switch">
-              <input id="iarxiv-recommender-toggle" type="checkbox" class="lab-toggle" aria-labelledby="label-for-iarxiv">
+              <input id="iarxiv-toggle"
+                     type="checkbox"
+                     class="lab-toggle"
+                     data-script-url="{{ url_for('static', filename='js/iarxiv.js') }}"
+                     aria-labelledby="label-for-iarxiv">
               <span class="slider"></span>
               <span class="is-sr-only">IArxiv recommender toggle</span>
             </label>
           </div>
           <div class="column lab-name">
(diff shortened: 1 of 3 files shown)


browse/templates/abs/labs_tabs.html
```diff
@@ -160,6 +160,21 @@ <h1>Recommenders and Search Tools</h1>
...
+          </div>
+          <div class="column lab-name">
+            <span id="label-for-iarxiv">IArxiv Recommender</span> <em>(<a href="/iarxiv_login">Email Signup</a> | <a href="https://iarxiv.org/about">What is IArxiv?</a>))</em>
+          </div>
+        </div>
+        {% endif -%}
+
       </div>
       <div style="min-height: 15px" id="connectedpapers-output"></div>
       <div id="coreRecommenderOutput"></div>
```
SBBCornell, line 180: I think were missing the div for the IArxiv output... and my understanding is that the output will be a link to the IArxiv login page on arxiv.org, plus a longer description. Since that output is so simple we might just build it as html right on this page instead of using a script to generate it like the other labs projects do. 

Do you think that sounds right @bmaltzan ?

bmaltzan: Hey @SBBCornell , yes, I didn't create a div for any output to be displayed, when the toggle is on. Sure, just a link to arxiv.org/iarxiv_login is need, and simple html that is toggled visible sounds ideal.

bmaltzan: I added the toggle. It's pretty simple, but I did end up adding it in a js script. It might be easier to read that way, just following the existing conventions. What do you think?

SBBCornell: looks good!

SBBCornell (approved):
Looks good! It would be great to push it all to beta for one last review.

bmaltzan:
Sure, the dev branch is now on beta, and iarxiv is working in the related-papers tab, eg: https://beta.arxiv.org/abs/2007.13745

PR #281, opened 2022-08-02
Fixes 500 errors abs file due to // for TeX cases

bdc34:
// at the start of a line for TeX cases was causing the abs parser to fail for https://arxiv.org/abs/2004.02153

https://arxiv.org/abs/2004.02153 works because Jake manually fixed the abs file.

This PR adds an alternative way to split the abs file to handle this case.

### DIFF
--- a/browse/services/document/metadata.py
+++ b/browse/services/document/metadata.py
@@ -490,6 +490,8 @@ class AbsMetaSession:
         # there are two main components to an .abs file that contain data,
         # but the split must always return four components
         components = RE_ABS_COMPONENTS.split(raw)
+        if len(components) > 4:
+            components = alt_component_split(components)
         if not len(components) == 4:
             raise AbsParsingException(
                 'Unexpected number of components parsed from .abs.')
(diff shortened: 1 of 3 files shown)


browse/services/document/metadata.py
```diff
@@ -490,6 +490,8 @@ def parse_abs_file(filename: str) -> DocMetadata:
         # there are two main components to an .abs file that contain data,
         # but the split must always return four components
         components = RE_ABS_COMPONENTS.split(raw)
+        if len(components) > 4:
+            components = alt_component_split(components)
```
bdc34, line 494: `alt_component_split()` is only used when the usual parse fails. This is to avoid introducing any bugs to the existing parse code.

tests/data/abs_files/ftp/arxiv/papers/2004/2004.02153.abs
```diff
@@ -0,0 +1,34 @@
...
+Categories: math.AP
+Comments: 16 pages
+MSC-class: 35K55 (primary), 35D99, 35Q92, 92C17 (secondary)
+License: http://arxiv.org/licenses/nonexclusive-distrib/1.0/
+\\
+  We construct global generalized solutions to the chemotaxis system
+  \begin{align*}
+  \begin{cases}
+  u_t = \Delta u - \nabla \cdot (u \nabla v) + \lambda(x) u - \mu(x) u^\kappa,
+\\
```
bdc34, line 19: Line 19 is the offending TeX case split.

DavidLFielding (approved):
Looks good. 

Are there any other places in our code that might be impacted by this bug?

bdc34:
> Looks good.
> 
> Are there any other places in our code that might be impacted by this bug?

Yes, this code is copy/pasted to arxiv-canonical.

PR #282, opened 2022-08-23
Upgrade to Python 3.10

bdc34:


.pylintrc
```diff
@@ -56,21 +56,10 @@ confidence=
 # --disable=W"
 disable=invalid-name,
         import-error,
```
bdc34, line 58: Newer versions of pylint no longer support these error-names.

Pipfile
```diff
@@ -6,51 +6,43 @@ name = "pypi"
 [packages]
 dataclasses = "*"
 python-dateutil = "*"
```
bdc34, line 8: Since arxiv-browse does not produce a package to upload to pypi I didn't convert to Poetry. 

I relaxed the dependency constraints.

browse/config.py
```diff
@@ -161,36 +163,6 @@
 available. This defaults to http.
 """
 
-JSON_AS_ASCII = os.environ.get("JSON_AS_ASCII") == ON
```
bdc34, line 164: These configs are no longer supported by Flask. It appears they were not in use anyway.

PR #290, opened 2022-10-27
ARXIVNG-5240: add cond-mat to iarxiv 2

bmaltzan:
Changes to startswith cond-mat


bmaltzan:
> Why is `starts-with` necessary?

cond-mat was split into cond-mat.supr-con, ...

DavidLFielding (approved):
Looks good.

PR #294, opened 2022-11-04
banner and news link to BPS

SBBCornell:
This commit adds a banner with link to the survey, and a news item on the homepage with a link to the same.

The start and end date on news.html should be updated to match whatever the start and end date of the banner are, so they appear at the same time. We want to display both for a 24 hour period. It is not so important exactly what time that 24 hour period begins, but preferably on Monday or Tuesday so we match previous years. 

Could we also preview it on dev or beta ahead of time, just to make sure all is well.


bdc34 (approved):

bmaltzan:
@SBBCornell , sure, it's here: https://dev.arxiv.org/

SBBCornell:
The banner on dev looks great!

PR #297, opened 2022-11-10
ARXIVNG-5256: institutional-banner

bmaltzan:
Create endpoint for looking up member institutions by ip address.
If found, cache for 30 days or until browser closes.
If not previously cached, and db is down, user won't see member acknowledgement until db is restored, but abs page will work if .abs file can be read from nfs.
Bots probably won't execute js, so may reduce number of queries.

### DIFF
--- a/browse/templates/base.html
+++ b/browse/templates/base.html
@@ -222,7 +222,7 @@
 
         expires_ms = 30*24*60*60*1000;
         if (lookup_failed) {
-          expires_ms = 4*60*60*1000; // try sooner if db issue.
+          expires_ms = 1*60*60*1000; // try sooner if db issue.
         }
         const new_expires = new Date();
         new_expires.setTime(new_expires.getTime() + expires_ms);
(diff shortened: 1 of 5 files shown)


SBBCornell (approved):
The description of the solution sounds good to me. I wont attempt to review the code, however.

browse/templates/base.html
```diff
@@ -184,6 +184,51 @@ <h2 id="mobilemenulabel">quick links</h2>
       </div>
     </footer>
   </div>
+
+  <script type="text/javascript">
+
+    function setCookie(cname, cvalue, exdays) {
```
bdc34, line 190: Can you do this with [local storage ](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) instead of a cookie?

bdc34 (changes requested):

browse/templates/base.html
bdc34: Also, can you move this to base so that all our projects can use this?

bdc34 (approved):

PR #298, opened 2022-11-11
ARXIVCE-62: abs-canonical-url

bmaltzan:
Some of our paper's abs pages aren't searchable at google. Maybe our multiple urls look like we're trying to game the page rank.


bdc34 (changes requested):
Thanks for getting on this.

The URL for the abs page should not need to be added to the `browse.config.URLS` because the abs page is a blueprint route in the arxiv-browse flask app. The URLS are just for URLs we'd like to use `url_for` with but do not exist in the arxiv-browse app as routes. 

Here is an example of getting the URL for the abs page: `url_for('browse.abstract', arxiv_id=abs_meta.arxiv_id)`

bdc34 (approved):

PR #301, opened 2022-11-17
labs pages: not default enabled for the demo tab, links for CORE and PwC

bdc34:
This undoes changes in 8dc1713 which made the demo tab visible and also enabled the replicate labs items in that tab for a subset of archives. It is an arXiv policy that the labs items are opt-in. 

This also has the effect of making the Demo tab enabled for all archives and categories. Previously it was just for cs, eess and stats.

Adds missing about links for CORE and PwC.

Adds main.py for easy startup of the flask app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ You can run the browse app directly. Using pipenv:
 
 ```bash
 pipenv install
-FLASK_APP=app.py FLASK_DEBUG=1 pipenv run flask run
+pipenv run python main.py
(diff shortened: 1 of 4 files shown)


SBBCornell (approved):
approve the updated 'about' links.
Not qualified to evaluate the enable toggles and tab changes.

jimentwood (approved):
Link changes look good. I'm not able to review the code changes.

PR #304, opened 2022-11-30
ARXIVNG-5256: remove pendo code

bmaltzan:
Disable geoip and remove analytics demos.

ie: 
arxiv.matomo.cloud
arxivcu.containers.piwik.pro


bdc34 (approved):
I think I might have removed the JS from the base.html in a HOTFIX.

bmaltzan:
Looks like there's still content.analytics.arxiv.org (pendo) in arxiv-docs, but that will go away with info.*
I don't see it anywhere else...

bmaltzan:
>

bmaltzan:
> I think I might have removed the JS from the base.html in a HOTFIX.

No problem

PR #307, opened 2022-12-15
adds og:image meta tag

SBBCornell:
Adds open graph meta tags so that facebook will use our logo instead of smileybones when scraping abs pages. 

~~There is still an 'inavlid image url' error that I dont know how to fix. [View the error here in the fb scraper](https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fbeta.arxiv.org%2Fabs%2F2007.09061)~~


SBBCornell:
these are the tags throwing an error on abs.html:

    <meta property="og:image" content="{{ url_for('static', filename='images/arXiv-logo-1200x700.png') }}"/>
    <meta property="og:image:secure_url" content="{{ url_for('static', filename='images/arXiv-logo-1200x700.png') }}" />

SBBCornell:
Removing numbers from the image name was part of the fix, though it is not anywhere in their documentation and seems like an insane policy, if it is one. It appears to be working now, at any rate:
Screen Shot 2022-12-16 at 9 26 42 AM

bdc34 (approved):
Looks good to me. Whoever deploys this needs to remember to deploy the static assets to the AWS CDN.

PR #316, opened 2023-03-06
Refactor listings to prepare for further changes

bdc34:
This refactors @DavidLFielding 's FS based listing code to prepare for further changes.

- Changes the API of listing service from returning TypedDict to returning classes
- Changes the controllers and templates to use the new classes
- Changes the listing implementations to use the new classes
- It uses cloudpathlib to open the listing file
- It moves the listing file parsing code to its own module
- combines some functions where if-modified-since and multiple listing files are handled

Much of the changes are in `fs_listing.py` and `parse_listing_file.py` but github is not showing those since they are "large diffs"


browse/services/documents/fs_implementation/legacy_fs_paths.py
```diff
@@ -12,7 +12,8 @@
 from .formats import list_ancillary_files
 from ..format_codes import has_ancillary_files, VALID_SOURCE_EXTENSIONS
 
-UPath = Union[CloudPath, Path]
+from browse.services import APath
```
bdc34, line 15: CloudPathLib has a function to_anyypath which returns either a Path for a file name like '/bobs/uncle' or a CloudPath for a name like 'gs://bobs/uncle'. This is a type to represent that.

browse/services/listing/__init__.py
```diff
@@ -1,32 +1,268 @@
...
+
+def fake(settings: Any, _: Any) -> "ListingService":
     """Factory function for fake listing service."""
     from .fake_listings import FakeListingFilesService
+
     return FakeListingFilesService()
+
+
+ListingTypes = Literal["new", "cross", "rep"]
+
```
bdc34, line 49: Below are the new listing classes that are class based versions of the previous TypedDict types. Classes with the attribute syntax are more pleasant to work with.

browse/services/listing/db_listing_impl.py
```diff
@@ -1,17 +1,16 @@
 """arXiv listings backed by a DB."""
```
bdc34, line 1: The DB implementation was changed to work with the refactor. But I'm thinking that we should go with the file based one since the DB one is incomplete and the tables it uses lack data for before 2010.

PR #323, opened 2023-03-10
Add specific versions of all modules in Pipfile .

jonathanhyoung:


bdc34:
It does not work well to pin the arxiv-browse Pipfile dependencies. The main problem is that if the arxiv-browse dependencies that are shared with arxiv-base or arxiv-auth are over constrained a solution to the dependency problem cannot be found. (Pinning dependencies that are not shared is not a problem, but the Pipfile.lock may be a better way to do it.)  If we pin everywhere we get this using flask an an example:

```mermaid
  graph TD;
      arxiv-browse-->arxiv-base;
      arxiv-browse-->arxiv-auth;
      arxiv-auth-->arxiv-base;

      arxiv-browse-->flask2.0.1;
      arxiv-base-->flask2.0.1;
      arxiv-auth-->flask2.0.1;
```
With all repos having pinned versions for flask, there needs to be a coordinated effort to keep the Pipfiles, and pipfile.locks all in sync. This was tried during the NG period and it lead to a lot of headaches.  Like
this:
```mermaid
  graph TD;
      arxiv-browse-->arxiv-base;
      arxiv-browse-->arxiv-auth;
      arxiv-auth-->arxiv-base;

      arxiv-browse-->flask1.4.8;
      arxiv-base-->flask2.2.1;
      arxiv-auth-->flask2.0.1;
```


This problem is repeated for every transitive dependency installed.

I've leaned toward having arxiv-base specify a relatively specific version of flask and then having the other arxiv repos unconstrained for the dependencies brought in via arxiv-base.

```mermaid
  graph TD;
      arxiv-browse-->arxiv-base;
      arxiv-browse-->arxiv-auth;
      arxiv-auth-->arxiv-base;
      arxiv-base-->flask2.0.1; 
```

bdc34:
Other ways to reduce the problems of handling these is to flatten the graph by combining repos. arxiv-auth could be moved into arxiv-base.

Another way would be to remove common dependencies from arxiv-base. Flask/werkzeug/jinja/markdown have been some of the libraries that cause problems. They are included in arxiv-base to support common web related features that we want to share across our web apps. If we want to remove a library like flask form arxiv-base we will have to create new techniques to achieve that.

bdc34 (changes requested):
Please let's discuss this, I'm open to full pinning if that is what we agree will work well.

PR #324, opened 2023-03-13
banner with Forum invitation

SBBCornell:
This PR includes edits to the banner and the home page with the forum invitation content. 

The banner in user_banner.html should run for 1 week from it's deploy date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@
 You can run the browse app directly. Using pipenv:
 
 ```bash
+python --version
+3.6.15
+python -m venv venv
+venv/bin/activate
+pip install pipenv==2022.4.8
 pipenv install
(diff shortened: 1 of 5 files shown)


bdc34 (approved):
I tried this and it works.

jonathanhyoung (commented):
Isn't {%- if 0 -%} a better choice than  ?  The template directive nests cleanly and is removed by template processing; the other will have problems if there are any comments in the code, and will appear in the source even when disabled.

bdc34:
> Isn't {%- if 0 -%} a better choice than ? The template directive nests cleanly and is removed by template processing; the other will have problems if there are any comments in the code, and will appear in the source even when disabled.

Good point.

PR #329, opened 2023-03-15
User banner fix

bdc34:
Changes the user banner so all of the control of the user banner lives in the `user_banner.html` file. 

Note this is a stacked PR that merges to `forum-banner` not to `develop`.


SBBCornell (approved):
that looks good to me!

PR #334, opened 2023-03-22
update forum messages

SBBCornell:
This small PR updates:

- the text about the forum on the homepage
- updates the forum banner text for later use (but not deploying now)
- and removes the "robots beware" text from the homepage


jimentwood (approved):
looks good to me

PR #335, opened 2023-03-22
Fixes home page link for CS

bdc34:
Set URL to CS as CoRR to info.arxiv.org and set description to archive/cs.

These URLs are currently broken on the home page.

https://arxiv-org.atlassian.net/browse/ARXIVCE-285


jweiskoff (approved):
looks good.

PR #343, opened 2023-04-26
Adds better error handling for /bibtex 

bdc34:
/bibtex/None is throwing 500 when it should be 400. Fixes this and some other cases.

When InternalServerError is explicitly raised the stack trace of the error does not end up in the logs. That hinders investigation of problems. 

ARXIVCE-339

### DIFF
--- a/tests/test_bibex.py
+++ b/tests/test_bibex.py
@@ -34,3 +34,28 @@ class BibexTest(TestCase):
             txt,
             "citation meta tag must have jref DOI",
         )
+
+    def test_good_bibtex(self):
+        rv = self.app.get(f"/bibtex/0906.3421")
+        self.assertEqual(rv.status_code, 200)
+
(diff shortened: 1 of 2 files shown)


tests/test_bibex.py
```diff
@@ -34,3 +34,28 @@ def test_bibex_supporting_meta_tags(self):
...
+
+        rv = self.app.get(f"/bibtex/0906.3421v1")
+        self.assertEqual(rv.status_code, 200)
+
+    def test_bibex_none(self):
+        """Don't do a 500 for /bibex/None ARXIVCE-339."""
+        rv = self.app.get(f"/bibtex/None")
+        self.assertEqual(rv.status_code, 400)
+
+    def test_bab_bibtex(self):
```
DavidLFielding, line 50: Do you mean 'bad' here?

bdc34: Oh, yes I did.

PR #345, opened 2023-05-07
Add microformats2 markup to arxiv

capjamesg:
This PR adds microformats2 [h-entry](https://microformats.org/wiki/h-entry) markup to Arxiv. h-entry markup makes it easy for software to parse the semantics of a page, widely used in the [IndieWeb community](https://indieweb.org/h-entry) and in large pieces of software such as the Tumblr default theme.

To test, one can run the markup through a microformats2 parser such as [mf2py](https://pypi.org/project/mf2py/) (Python).

For more information on microformats:

- [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/microformats)

### DIFF
--- a/browse/templates/abs/abs.html
+++ b/browse/templates/abs/abs.html
@@ -41,7 +41,7 @@
 
   <div class="leftcolumn">
     <div class="subheader">
-      <h1>{% if abs_meta.primary_archive.id != abs_meta.primary_category.id %}{{ abs_meta.primary_archive.name }} > {% endif %}{{ abs_meta.primary_category.name }}</h1>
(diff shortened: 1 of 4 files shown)


browse/templates/abs/abs.html
```diff
@@ -41,7 +41,7 @@
 
   <div class="leftcolumn">
     <div class="subheader">
-      <h1>{% if abs_meta.primary_archive.id != abs_meta.primary_category.id %}{{ abs_meta.primary_archive.name }} > {% endif %}{{ abs_meta.primary_category.name }}</h1>
```
cbf66, line 44: Why the change from <h1> to <p> here?

That could have a negative affect on screen readers.

PR #347, opened 2023-05-10
Fixes bug in withdrawn msg on abs page

bdc34:
The name on the withdrawal message reflects the submitter of the withdrawal. Previously it was the submitter of the version you were viewing. So if v7 was wdr by admins but you were looking at v1 it would have the a withdrawal message with the wrong name.

image


jweiskoff (approved):
Looks good. Once we have version-level withdrawals this will also make sense to phrase this way, but I think it's probably good enough for now.

PR #348, opened 2023-05-11
Changes withdrawn abs page

bdc34:
Red boarder to black border. 

Withdrawn is not an error state. 

Removes download link to source ( https://arxiv.org/e-print/{arxiv_id} ) since there is not any source.

Bolds comment. Jim asked for a box similar to the withdrawn message but that doesn't work due tot he comment being in a table.

image

### DIFF
--- a/browse/controllers/abs_page/__init__.py
+++ b/browse/controllers/abs_page/__init__.py
@@ -140,13 +140,9 @@ def get_abs_page(arxiv_id: str) -> Response:
                     response_data["higher_version_withdrawn"] = True
                     response_data["higher_version_withdrawn_submitter"] = _get_submitter(abs_meta.arxiv_identifier, ver_index+1)
 
+        response_data["withdrawn"] = abs_meta.version_history[abs_meta.version - 1] in response_data["withdrawn_versions"]
 
-        # Following are less critical and template must display without them
-        # try:
         _non_critical_abs_data(abs_meta, arxiv_identifier, response_data)
-        # except Exception:
-        #    logger.warning("Error getting non-critical abs page data",
-        #                   exc_info=app.debug)
 
     except AbsNotFoundException:
         if (
--- a/browse/templates/abs/abs.html
+++ b/browse/templates/abs/abs.html
@@ -230,8 +230,7 @@
 {% endif %}
 <div id="content-inner">
   <div id="abs">
-      {%- set current = abs_meta.version == abs_meta.version_history[-1].version -%}
-      {% if abs_meta.version_history[abs_meta.version - 1] in withdrawn_versions %}
+      {% if withdrawn %}
           <span class="error" style="border: 2px solid grey">This paper has been withdrawn by {{ abs_meta.submitter.name|tex2utf if abs_meta.submitter.name != None }}</span>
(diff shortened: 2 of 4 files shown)


jweiskoff:
Does this also remove the option to go to `arxiv.org/e-print/[id]`? If so, this could solve several other problems we have.

bdc34:
> Does this also remove the option to go to `arxiv.org/e-print/[id]`? If so, this could solve several other problems we have.
Yes this removes that link and replaces it with the word "withdrawn".

jimentwood (approved):
Looks good from the screenshot for the current scope of changes.
Thanks!

jweiskoff:
> > Does this also remove the option to go to `arxiv.org/e-print/[id]`? If so, this could solve several other problems we have.
> 
> Yes this removes that link and replaces it with the word "withdrawn".

That's not what I mean. Will it prevent that url from working? Currently, admins have to create an essentially blank `withdrawn.txt` file to replace the contents of a work, and remove the complete set of files to prevent crawlers from still being able to download withdrawn files. This PR probably doesn't address that issue, but this will remove some of the confusion from our users as-is.

jimentwood:
@jweiskoff I don't think that was in scope. Dropping the displayed link, should not remove the existing url. If there are programmatic users who fetch files they will not be affected by the display change....
right @bdc34 ?

chrisjcameron:
My best understanding is that this uses information from the ArXiv database to detect which version of the paper was withdrawn and to 1) prevent links to withdrawn versions from displaying; 2) display the withdrawn message on the withdrawn version(s); 3) display warnings on any version prior to the withdrawn version (No warnings for later versions of previously withdrawn papers). This PR does not address how papers are assigned the withdrawn status. A feature that removes withdrawn files completely should be implemented at a different stage of the withdrawal process, probably about the same time that withdrawal is entered into the database.

jimentwood:
Hi Christopher,

Yes, that is my understanding of the scope. The current changes will help readers avoid some of the common confusions and head off questions we get to the support desk.

It may be good to chat before doing future work. For example we don't want any files completely removed for withdrawn submissions.

There does need to be a totally new status for "Removed" submissions, and a whole new backend/frontend for administrators to support that, as well as changes to abs display. The new metadata model includes a proposed list of "posting status" options of which withdrawn and removed are part of https://github.com/arXiv/arxiv-canonical/blob/4f8bdbffcb72161bddfb1ccab6431f089a2b04b8/schema-extended/Common.json#L49

There is also a plan to indicate a version is withdrawn in the new metadata model.
That would clear up ambiguity in the current model which creates a new version that overlays all previous versions. The current model is not ideal, for example, if someone submits v3 with an error and then corrects it in v4.

The new model would allow a version posted to be withdrawn and the other versions not to be withdrawn.

At some point we should describe the ideal states so that changes can work in that direction. It may be good to get some UX research around this area as well so it isn't just my view, which tends to be influenced by problems users encounter rather than by what researchers ideally need.

Best,
Jim
________________________________

bdc34:
@jweiskoff  This PR does not change the /e-print code. It will still serve up the dummy file. I wouldn't be surprised if the dummy file is also needed to get the publish to go smoothly for a wdr.

PR #350, opened 2023-05-22
Hiring banner

SBBCornell:
adds minimal time-boxed banner to header with hiring link. It does not use our traditional banner code. This banner cannot be closed, but it also does not add to the header height or cover up any other content. It is set to stop displaying on August 1st, but this date can easily be adjusted.

Here are two color options for red-themed (current):
Screen Shot 2023-05-22 at 11 36 29 AM

and blue-themed:
Screen Shot 2023-05-22 at 11 39 45 AM

### DIFF
--- a/browse/static/css/arXiv.css
+++ b/browse/static/css/arXiv.css
@@ -655,6 +655,28 @@ footer .sorry-app-links .a11y-main-link {
     width: 88%;
   }
 }
+.banner-minimal {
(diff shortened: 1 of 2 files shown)


SBBCornell:
Updated to a more minimal button style:
Screen Shot 2023-05-22 at 3 02 06 PM
@jimentwood

PR #358, opened 2023-06-10
Deploy "we're hiring" banner to production!

cbf66:


bdc34:
There is a file named `git` in the root of this project.  Please remote this.

bdc34 (approved):
Looks good.  Please remove `git`.

Since this has css changes the css file will need to be deployed to the CDN. Let' me know if you want me to do it for you. See https://github.com/arXiv/arxiv-browse#serving-static-files-on-s3

browse/templates/home/special-message.html
```diff
@@ -1,20 +1,21 @@
 {#- Special messages appear in a column to the side of the website intro/news section,
   each in their own column. Always apply the style "message-special" to the column.
   You can also apply the style "dark" to override the red styles and add a message with a black background instead.
-  There should preferably be just one, and never more than two. -#}
+  There should can be just one, never more than two. -#}
```
jonathanhyoung, line 4: You should can this message in favor of one with better grammar...

jonathanhyoung (approved):

browse/templates/home/special-message.html
cbf66: It's just a comment ...

PR #360, opened 2023-06-14
Html conversion for browse

mnazzaro:
The HTML feature seems to be working well on browse


bdc34:
This work was done in other PRs.

PR #370, opened 2023-07-11
Update README.md

1Haschwalth:
Heading2


jonathanhyoung:
This change is not needed and is rejected.  We have a robust ticketing process (mostly in JIRA) if you are interested in contributing.

PR #372, opened 2023-07-18
remove broken links (again)

mnazzaro:


bmaltzan:
Hey @mnazzaro , we want these urls to be in production on the cit vms, so let's talk about another option.

jonathanhyoung (approved):

PR #408, opened 2023-10-12
adds minimal banner and homepage block for prod editor role

SBBCornell:
This PR adds two notices related to the open role for Production Editor: 
- A minimal banner in the header
- and a homepage block

Reviewers: please double check the start and end dates set for these banners to appear: 
if rd_int >= 202310120100 and rd_int <= 202311150100
so from 10/12 at 1:00am to 11/15 at 1:00am.

Also, please double check that the url I am linking to is the one we want to use: [workday job page](https://cornell.wd1.myworkdayjobs.com/en-US/CornellCareerPage/job/arXiv-Production-Editor--Cornell-Tech--NYC-_WDR-00040646)

See screenshot:
Screen Shot 2023-10-12 at 10 21 23 AM

### DIFF
--- a/browse/templates/base.html
+++ b/browse/templates/base.html
@@ -34,12 +34,17 @@
         <div class="column" id="cu-logo">
           <a href="https://www.cornell.edu/"><img src="{{ url_for('static', filename='images/icons/cu/cornell-reduced-white-SMALL.svg') }}" alt="Cornell University" /></a>
         </div>
-        {%- if config["BROWSE_MINIMAL_BANNER_ENABLED"] and rd_int >= 202310010100 and rd_int <= 202312300100 -%}
-        <div class="column">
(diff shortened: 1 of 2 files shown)


bmaltzan:
- The date format is good.
- The job link is the public one, that's good.
- We're using that space, for the "Served from the cloud" message: https://browse.arxiv.org/

Locally, this would enable any messages that are toggled:
```
BROWSE_MINIMAL_BANNER_ENABLED=1 BROWSE_SPECIAL_MESSAGE_ENABLED=1  BROWSE_USER_BANNER_ENABLED=1 python main.py
```

jimentwood (approved):

SBBCornell:
> * We're using that space, for the "Served from the cloud" message: https://browse.arxiv.org/
>
Here is a screenshot of what it will look like if both messages are displayed at the same time:
Screen Shot 2023-10-12 at 12 25 42 PM

bmaltzan (approved):

PR #409, opened 2023-10-12
Fixes a bug in sync.

bdc34:
Removes bug in debugging statement.


DavidLFielding (approved):
Looks like the log entry that is generating the error is not needed. OK.

PR #414, opened 2023-10-18
Adds breaks and abs to fmt=txt

bdc34:
Jake pointed out that this was not the correct format.

ARXIVCE-701


jweiskoff (approved):
looks good.

PR #415, opened 2023-10-19
adds giving week banner

SBBCornell:
This PR updates user_banner.html with the giving week banner content, and updates the display date range from Monday 10/23 through the end of the day of 10/29 (I set the end date to 10/30, is that right? or should it be 10/29?).

### DIFF
--- a/browse/templates/user_banner.html
+++ b/browse/templates/user_banner.html
@@ -24,6 +24,7 @@ Adding an environment var, so we can enable only on gcp.
 
 
 {%- macro content(request_datetime) -%}
+{%- if (BANNER_START <= request_datetime.strftime('%Y%m%d')) and (BANNER_END > request_datetime.strftime('%Y%m%d')) %}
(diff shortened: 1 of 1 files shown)


bmaltzan:
Hey Shamsi,
 
Yes, I think 10/30 is the right one. The time should default to midnight. You could test by setting BANNER_END to today and tomorrow. And BANNER_START to yesterday. 

This flag was recently created so we could enable the "cloud" messages only in https://browse.arxiv.org/ and not https://arxiv.org/

I think you'll want to remove the two conditions for the aside and slider.css:
config['BROWSE_USER_BANNER_ENABLED']

The banner didn't show up for me locally, until I removed the style="display:none.
I'm loading http://127.0.0.1:8080/static/browse/0.3.4/css/slider.css?v=1.1
which may not be right.

bmaltzan (approved):
Cool, looks like you found some css.
I set an end date, so we don't have to deploy next week.

PR #419, opened 2023-10-23
Arxivce 805 html route

mnazzaro:


browse/config.py
```diff
@@ -116,7 +118,7 @@ class Settings(BaseSettings):
     - `browse.services.listing.fake`: A totally fake set of listings for testing.
     """
 
-    DOCUMENT_LISTING_PATH: str = 'tests/data/abs_files/ftp'
+    DOCUMENT_LISTING_PATH: str = 'gs://arxiv-production-data/ftp'
```
bdc34, line 121: Do not change these defaults to GS buckets as it would mess up the tests.

browse/domain/metadata.py
```diff
@@ -77,6 +77,8 @@ class DocMetadata:
     submitter: Submitter
     """Submitter of the article."""
 
+    source_type: Optional[str] = None
```
bdc34: Please do something like this in the pacakge:
`source_type = Literal["tex", "ps", "html" .etc ]`
and then 
`source_type: Optional[source_type] = None`
Possibly a better name for the literal type than source_type?

bdc34 (changes requested):
One small change related to the source_type, and maybe using a `Literal`.

PR #428, opened 2023-11-06
Sync script changes

mnazzaro:


browse/static/bibex/bibex.js
bdc34, line 1: It doesn't seem related to the main changes. Please remove this change.

browse/static/bibex/bibex.min.js
bdc34, line 1: It doesn't seem related to the main changes. Please remove this change.

browse/templates/home/home.html
bdc34, line 1: It doesn't seem related to the main changes. Please remove this change.

bdc34 (changes requested):
Seems good but there are thee files that are unrelated that changed.

PR #436, opened 2023-11-27
Update cite.py to include arXiv url and doi

daringli:
Many journal bibtex style (.bst) files do not support arXiv IDs. By adding doi and url fields, the preprint bibtex export becomes compatible with these .bst files.
I have not tested these changes since I'd have no idea how to run the flask app. Please make or suggest any edits needed if it doesn't work.


daringli:
Feature got added

PR #438, opened 2023-11-30
HTML-UI: light and dark mode fixes

zzhilin:


jweiskoff:
This has been hanging out since November. Is it still viable? if not can we close it?

PR #443, opened 2023-12-19
Arxivce 646 speed up year path

kyokukou:


kyokukou:
wasnt sure if I should commit the change to the config file, but none of this will run if not set use the hybrid listing system

tests/conftest.py
```diff
@@ -133,6 +152,10 @@ def client_with_fake_listings(app_with_fake):
     with app_with_fake.app_context():
         yield app_with_fake.test_client() # yield so the tests already have the app_context
 
+@pytest.fixture(scope='function')
```
bdc34, line 155: This seems like it could be session scope.

bdc34 (approved):

PR #445, opened 2023-12-21
deploys funny paper id changes

bdc34:


tests/conftest.py
```diff
@@ -21,12 +21,18 @@
...
 
+import browse.services.documents as documents
+from browse.services.listing import db_listing
+
 DEFAULT_DB = "sqlite:///../tests/data/browse.db"
 TESTING_LATEXML_DB = 'sqlite:///../tests/data/latexmldb.db'
 
 
 TESTING_CONFIG = {
     "SQLALCHEMY_BINDS": {"latexml": TESTING_LATEXML_DB},
```
kyokukou, line 32: this leaves changes to latexmldb.db when I run tests.

tests/conftest.py
```diff
@@ -21,12 +21,18 @@
...
+    "SQLALCHEMY_DATABASE_URI" : DEFAULT_DB,
+    'DOCUMENT_LISTING_SERVICE': db_listing,
```
kyokukou, line 34: i dont think this is fully/ properly implemented. I believe tests either use the fake listing service or the file listing service with the listing file path pointing to local test data

tests/conftest.py
bdc34: Yes it does. We'll fix that soon.

kyokukou (approved):

PR #446, opened 2023-12-21
adds . to end of just math archive request

kyokukou:


kyokukou:
/year sorts for what archive an article belongs in by seeing if the categories feild of the metadata table contains the string of the archive name. A ".' was added to the end of the archive name to separate "math" from "math-ph". This resulted in archives without subcategories finding no entries. This updates to just add the period for the math special case which is the only archive name that also starts another archive name.

bdc34 (approved):

PR #448, opened 2023-12-21
Set abs page cache-control to 1h

bdc34:
As requested by admin team to allow smoother support.


jimentwood (approved):
I'm not reviewing the code, but the concept sounds good for now

PR #450, opened 2023-12-22
years_operating now alwasy tries to add just the year of dates

kyokukou:
fixes a bug for archives that have end years where it would try and add a date and an int together


bdc34 (approved):
Nice

PR #462, opened 2024-01-23
Fix issue with shadowing var files, fix url to ms in logs

mnazzaro:


mnazzaro:
Fixing this ticket: https://arxiv-org.atlassian.net/browse/ARXIVCE-1174

ntai-arxiv (approved):
This would make it better but can you come up with a test? there is a test/test_sync_prod_to_gcp.py which has some scaffolding done.

mnazzaro:
Seems tricky to test

PR #470, opened 2024-01-30
Range request handling for /src and /e-print

bdc34:
**Adds range request handling for both `/src` and `/src/{paperid}/anc/{path}`**
Adds range request handling and tests.

**Changes behavior of /src**
This changes `/src/{paperid}` to download the source in the format that it exists in the storage. This is a behavior change that affects users.

The reasons for this change:

1. fastly and other CDNs need to have Range Requests and partial content enabled to handle large files (> 20MB) and Content-Length

2. We want to use streaming to avoid copying large files to the limited storage on containers.

3. The final Content-Length will not be known when streaming a response out via a pipe of tar -> gzip

**Old behavior described below this line:**
The path /src use to always download a .tar.gz file. It would always set the encoding to gzipped.

In the common case of .tar.gz it just sent the file.

In the case of a single file like .pdf it tarred and gziped that.

In the case of a .gz single file like a .ps.gz it would unzip, tar and regzip.


browse/controllers/files/__init__.py
```diff
@@ -1,9 +1,12 @@
-from typing import Iterator
+from typing import Iterator, Union
 from email.utils import format_datetime
 from flask import Response
 from datetime import timezone
+import mimetypes
 
+from browse.domain.identifier import Identifier
 from browse.domain.metadata import Identifier
```
jonathanhyoung, line 8: This is unlikely to be correct - ?

browse/services/object_store/fileobj.py
```diff
@@ -37,7 +59,7 @@ def exists(self) -> bool:
         pass
 
     @abstractmethod
-    def open(self,  *args, **kwargs) -> IO:  # type: ignore
+    def open(self, mode:str) -> BinaryMinimalFile:
```
jonathanhyoung, line 62: "pass" does not return a BinaryMinimalFile - what type do you really want here?

browse/services/object_store/fileobj.py
```diff
@@ -187,10 +209,10 @@ def name(self) -> str:
...
         return self._fileobj.exists()
 
-    @contextmanager
-    def open(self, *args, **kwargs) -> IO:  # type: ignore
-        with self._fileobj.open(mode="rb") as zipped:
-            yield gzip.GzipFile(filename="", mode="rb", fileobj=zipped)
+    def open(self, mode:str) -> BinaryMinimalFile:
+        return gzip.GzipFile(filename="",
+                             mode=mode,
+                             fileobj=self._fileobj.open(mode))
```
jonathanhyoung, line 215: Again, I'm not sure what return type you want here, but BinaryMinimalFile seems wrong.

browse/services/object_store/fileobj.py
```diff
@@ -205,10 +227,11 @@ def size(self) -> int:
...
             # the whole file.
-            with self.open(mode="rb") as unzipped_fh:
-                size = unzipped_fh.seek(0, io.SEEK_END)
-                self._size = size
-                return self._size
+            unzipped_fh = self.open(mode="rb")
+            size = unzipped_fh.seek(0, io.SEEK_END)
+            unzipped_fh.close()
+            self._size = size
+            return self._size
```
jonathanhyoung, line 234: Why did this change?  The "with ..." wraps the indented block with a try / finally close, and the new code is not similarly protected against thrown errors.

jonathanhyoung (approved):
Please see my comments.

browse/services/object_store/fileobj.py
bdc34: BinaryMinimalFile is a protocol that I'm using to inform Mypy of the minimal interface I need to get range requests to work.
At this line mypy should be saying "Okay it's returning BinaryMinimalFile, does the object returned meet that interface?"

browse/services/object_store/fileobj.py
bdc34: `IO` is a crazy pile of signatures. It didn't seem usable with mypy. So I made a minimal interface via a `Protocol` to make things easier.

PR #471, opened 2024-01-30
Avoids assertion error when latexml is not enabled.

bdc34:


mnazzaro:
Why are we checking in the database service? I think we should check earlier

mnazzaro (approved):

PR #472, opened 2024-01-30
[ARXIVCE-1176] Use genpdf-api from browse PDF.

ntai-arxiv:


.gitignore
```diff
@@ -119,7 +119,7 @@ celerybeat.pid
 *.sage.py
 
 # Environments
-.env
```
bdc34, line 122: Someone probably has a .env they would like still ignored. Please leave this line in.

browse/services/dissemination/article_store.py
```diff
@@ -375,6 +393,58 @@ def _pdf(self, arxiv_id: Identifier, docmeta: DocMetadata, version: VersionEntry
                      [str(ps_cache_pdf), str(pdf_file)])
         return "UNAVAILABLE"
 
+    def _genpdf(self, arxiv_id: Identifier, docmeta: DocMetadata, version: VersionEntry, api: str) -> FormatHandlerReturn:
+        """Gets a PDF from the genpdf-api."""
+        if version.source_flag.cannot_pdf:
```
bdc34, line 398: This clause is redundant, already handled by `_pdf()`

browse/services/dissemination/article_store.py
```diff
@@ -375,6 +393,58 @@ def _pdf(self, arxiv_id: Identifier, docmeta: DocMetadata, version: VersionEntry
...
+                logger.error("??? %s", bucket_url)
+                raise NotImplementedError("Somthing is wrong here")
+
+            case http.HTTPStatus.OK:
+                # Sadly, if you get 200, error out as this is not expected
+                logger.error("genpdf-api should not return 200. Check URL and turn off download")
+                raise NotImplementedError("Cannot support getting PDF")
+
+            case http.HTTPStatus.REQUEST_TIMEOUT:
+                logger.error("genpdf-api request timed out")
```
bdc34, line 435: Might be useful to add `t_end` to error message.

browse/services/dissemination/article_store.py
```diff
@@ -375,6 +393,58 @@ def _pdf(self, arxiv_id: Identifier, docmeta: DocMetadata, version: VersionEntry
...
+
+        match response.status_code:
+            case http.HTTPStatus.FOUND:
+                bucket_url = response.headers.get('location')
+                if not bucket_url:
+                    logger.error("Redirect did not provide location")
+                    return "UNAVAILABLE"
+                if isinstance(self.objstore, GsObjectStore):
+                    bucket_name, obj_key = from_bucket_url_to_key(bucket_url)
+                    if obj_key and bucket_name == self.objstore.bucket.name:
```
bdc34, line 424: This seems to require that the bucket sent back by genpdf-api is the same as the one configured in arxiv-browse. Is that going to always be the case? Does the flask app need a config of "GENPDF_API_BUCKET" and then have a genpdf_objstore?  

`FormhandlerReturn` can be a `FileObject` and google.cloud.stroage `Blob` is a `FileObject`

bdc34 (changes requested):
Looks good. 

The tests need to pass.

browse/services/dissemination/article_store.py
ntai-arxiv: I am not clear on this TBH. GenPDF's location is the full bucket path and that's fine. Whether or not Browse uses same bucket is probably Browse's business. Here, I basically chickened out to say that they should use the same bucket but they don't have to. 

Having separate bucket store may make sense. In some sense, it doesn't hurt to do so, so I'll do it. If we change mind, not a big deal.

> FormhandlerReturn can be a FileObject and google.cloud.stroage Blob is a FileObject

Not quite sure on this. I'll ask you in Slack.

bdc34 (approved):
Nice!

PR #477, opened 2024-02-02
Add test for html_conversion_dissemination

mnazzaro:


tests/test_db_abs.py
```diff
@@ -69,3 +69,13 @@ def test_db_abs_null_source_size(dbclient):
...
     download_button_html = html.select_one(".download-html")
     assert download_button_html is None
+
+def test_html_conversion_dissemination (dbclient):
+    rt = dbclient.get('/abs/0906.2112')
+
+    assert rt.status_code == 200
+    
+    html = BeautifulSoup(rt.data.decode('utf-8'), 'html.parser')
+    assert html.select_one('.extra-services') \
```
bdc34, line 79: I like to go directly for the a tag like:

       atag = html.select_one('.extra-services').find('a', {'href': '/html/physics/9707012v'})
        assert atag and atag.txt == "HTML (experimental)"    

Because if the HTML structure change this test will still pass. Other times you really want the <a> tag in a specific structure so do the whole path but I don't think the .full-text .ul or .li[1] are critical in this case.

bdc34 (approved):

PR #485, opened 2024-02-14
Change /abs to use etag instead of last-modified

bdc34:
We lack a good record for all the data sources that makes up the /abs page so last-modified isn't working out.

Go to etag made from a hash of all the data used by the jinja html template.


mnazzaro (approved):
I don't think the html tests will do much anymore but looks good to me

PR #487, opened 2024-02-14
develop => master

bdc34:
Adds src to /abs download area.

Fixes bug related to /src


mnazzaro (approved):
the html changes and showing src on abs looks good to me

PR #497, opened 2024-02-26
Arxivce 1070 monthly listings from db

kyokukou:
changes browse to use listings from the gcp database. also changes /list paths to use 4 digit years in URLs - note that this is incompatible with legacy system which uses 2 digit years. this means /year /list and /archive paths all need to be run from browse or not together. This PR leaves the old url for month in the browse abs pages, so it will currently be compatible with legacy, but the change will need to be made to a 4 digit year to serve the /list paths from browse


browse/controllers/list_page/__init__.py
```diff
@@ -113,13 +140,23 @@ def get_listing(subject_or_category: str,
...
+        not subject_or_category or
+        not (
+            time_period and
+             (time_period.isdigit()
+              or time_period in ['new', 'current', 'pastweek', 'recent']
+              or year_month_pattern.match(time_period)
+              )
+        )
+    ):
+        raise BadRequest #TODO make a less dramatic error
```
bdc34, line 156: 400 seems like the correct error.

bdc34 (approved):

PR #498, opened 2024-02-26
Changes /pdf to save in file named {paper_id}v{version}.pdf

bdc34:
At some point the legacy system was changed to use a PDF download URL like /pdf/{paper_id}.pdf and then to do redirects from /pdf/{paper_id} to that. This was solely to get the file downloaded to a reasonable name.

This commit uses a Content-Disposition header to get a reasonable download name.

It removes the /pdf/paperid -> /pdf/paeprid.pdf redirect.

It also changes the .pdf URLs to redirect to non .pdf URLs.


browse/controllers/files/dissemination.py
```diff
@@ -83,27 +85,38 @@ def src_resp_fn(format: FileFormat,
...
 
+def pdf_resp_fn(format: FileFormat,
+                 file: FileObj,
+                    arxiv_id: Identifier,
+                    docmeta: DocMetadata,
+                    version: VersionEntry,
+                    extra: Optional[str] = None) -> Response:
+    """funciton to make a `Response` for a PDF."""
+    resp = default_resp_fn(format, file, arxiv_id, docmeta, version, extra)
+    filename = f"{arxiv_id.filename}v{version.version} .pdf"
```
ntai-arxiv, line 107: Extra space befor .pdf is probably not what you want.

bdc34: Thanks!

jimentwood:
I am unsure what the goal or problem this is addressing.
Could we possibly talk it through?
I am also curious about user support impacts, if any.

________________________________

browse/controllers/files/dissemination.py
```diff
@@ -83,27 +85,38 @@ def src_resp_fn(format: FileFormat,
...
     add_time_headers(resp, file, arxiv_id)
     return resp  # type: ignore
 
+def pdf_resp_fn(format: FileFormat,
+                 file: FileObj,
+                    arxiv_id: Identifier,
+                    docmeta: DocMetadata,
+                    version: VersionEntry,
+                    extra: Optional[str] = None) -> Response:
+    """funciton to make a `Response` for a PDF."""
```
norbusan, line 105: function

browse/routes/dissemination.py
```diff
@@ -7,44 +7,38 @@
...
 blueprint = Blueprint('dissemination', __name__)
 
 
-@blueprint.route("/pdf/<string:archive>/<string:arxiv_id>", methods=['GET', 'HEAD'])
-@blueprint.route("/pdf/<string:arxiv_id>", methods=['GET', 'HEAD'])
+@blueprint.route("/pdf/<string:archive>/<string:arxiv_id>.pdf", methods=['GET', 'HEAD'])
+@blueprint.route("/pdf/<string:arxiv_id>.pdf", methods=['GET', 'HEAD'])
 def redirect_pdf(arxiv_id: str, archive=None):  # type: ignore
-    """Redirect urls without .pdf so they download a filename recognized as a PDF."""
+    """Redirect urls without .pdf to a URL to download a filename recognized as a PDF."""
```
norbusan, line 20: Redirect urls **with** .pdf ....

bdc34:
> I am unsure what the goal or problem this is addressing. Could we possibly talk it through? I am also curious about user support impacts, if any.

This is re-addressing the "When I click arXiv's 'download PDF' I want it saved in a file ending with .pdf so when I open it later on my computer, my computer knows to open it with a PDF viewer." This was addressed quite a while ago by redirecting from /pdf/{paperid} to /pdf/{paperid}.pdf. 

The redirect worked. But there is a standard mechanism that is part of HTTP to provide a way for the server to specify what the file name should be on download. It is called a content-disposition header. I think we started using the redirect because of an ignorance of the "inline" mode of the content-disposition header. 

This change preservers the "if I download this to my computer it is named {paper_id}.pdf" behavior and the content-disposition header in the response directly communicates with standards defined mechanisms that we want that behavior. It also eliminates the redirect, which is wasteful and only indirectly communicates what we want to happen.

There should be no functional impact on users of the web site. The impact facilitates an improved experience of the expected behavior. When they click the download link and get the PDF displayed in the browser. When the hit save inside the browser PDF viewer the file will be saved with the .pdf extension. The URL when looking at the PDF in the browser it will not have a .pdf it will be like https://arxiv.org/pdf/2402.12345 as it use to be.

The user should not notice the removal of the redirect.

To support any user who has bookmarked or saved a PDF URL with the .pdf extension will get a redirect to the "normal" PDF URL. 

If you have any additional questions let me know. If you want to meet about this also let me know.

jimentwood:
Thanks @bdc34 ! I was not sure what I was being asked to review initially. The changes all sound good. I am not reviewing the code itself. If you need an extra set of eye on testing let me know.

bdc34:
> Thanks @bdc34 ! I was not sure what I was being asked to review initially. The changes all sound good. I am not reviewing the code itself. If you need an extra set of eye on testing let me know.

I'll try to add a specific note to you in the future.

And thanks, the questions you brought up were exactly the kinds of thinks I would like to hear about from you.

ntai-arxiv (approved):

PR #499, opened 2024-02-26
Reduces the DB calls to get the DocMetadata for an abs page

bdc34:
This removes a unnecessary DB call when getting the DocMetadata object for a paper which is used in /abs. 

Before there was a DB query first for the specific version paper, then a DB query for all versions of a paper. The improvement is to make a single query for all versions.

convert.py was merged to db_abs.py since it was tightly coupled with the db_abs code.


DavidLFielding (approved):
Looks good. 

Any measure of potential improvement with cutting DB calls in half for the DocMetadata service? Maybe a comparison of large number of requests between old and new implementation. I'm just curious.

bdc34:
> Looks good.
> 
> Any measure of potential improvement with cutting DB calls in half for the DocMetadata service? Maybe a comparison of large number of requests between old and new implementation. I'm just curious.

I profiled it on my laptop connecting to the DB at GCP. Before this change the call to db_abs._abs_for_version() was 0.141s and after it was 0.077s

PR #501, opened 2024-02-27
Get large PDFs working with fastly

bdc34:
Refactors dissemination.py so most `responses` take advantage of the default response to get consistent behavior for large objects at GCP Cloud Run and Fastly's CDN.

Fastly needs range response to do "segmented caching" to handle large responses. GCP cloud run needs either Range requests or chunked encoding to do large responses. The current master does not do a chunked response, it expects to always get a range request. Some how fastly is doing non-range requests for the first GET and GCP will respond with a 500.

Moves some functions from dissemination.py to files/__init__.py just to keep that file shorter.

Removes file_processing.py, no longer used.


DavidLFielding (commented):
The range versus default chunked response makes sense and look good. Refactoring looks fine to me. 

Are tests needed for new Transform objects (in fileobj.py)?

bdc34:
> The range versus default chunked response makes sense and look good. Refactoring looks fine to me.
> 
> Are tests needed for new Transform objects (in fileobj.py)?

I've added a test.

PR #510, opened 2024-03-01
change button text for PDF and HTML

SBBCornell:
This PR is basically a very minor text change, but I am seeking feedback before it is merged because it is a visual change on abs.

At the NSF meeting we received feedback to change the "Download PDF" button text to "View PDF", and to update the HTML link to match. In order for the HTML text not to wrap to 2 lines I need to widen the sidebar a tad. Here are screenshots on desktop and mobile:

Desktop:
Screenshot 2024-03-01 at 2 15 50 PM

Mobile:
Screenshot 2024-03-01 at 2 18 13 PM

### DIFF
--- a/browse/static/css/arXiv.css
+++ b/browse/static/css/arXiv.css
@@ -887,7 +887,7 @@ p.tagline {
 #abs-outer .leftcolumn {
   margin: 0 0 1em 0;
   padding: 0px;
-  width:calc(100% - 16em);
+  width:calc(100% - 18em);
   float:left;
 }
 #abs-outer .mobile-submission-download {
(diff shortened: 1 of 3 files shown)


chrisjcameron:
Is the verb necessary? Could it just read "PDF", "HTML" and "Tex Source". I am not sure what is considered best practice but the behavior of the PDF button depends on how the user's browser is configured to handle incoming PDF files, so it opens in a viewer for some users and downloads for others.

cbf66:
From Paul: 

> i like the "view ..." for both.
> but instead of 'experimental' it should use the more conventional 'beta', i.e., "view html (beta)"
> which also has the significant advantage that it's shorter and won't cause the linewrap.

SBBCornell:
> From Paul:
> 
> > i like the "view ..." for both.
> > but instead of 'experimental' it should use the more conventional 'beta', i.e., "view html (beta)"
> > which also has the significant advantage that it's shorter and won't cause the linewrap.

We debated between 'beta' and 'experimental' originally. We landed on 'experimental' because beta is not as easily understood by arXiv's international audience.

bdc34 (approved):
I very much like the change to "View PDF". :+1:

PR #511, opened 2024-03-01
reworded 400 response to make sense for bad request

kyokukou:
this template is the default for user requests that are incorrect in some way


alisonhofer:
Approved!

PR #518, opened 2024-03-06
fix stats 500 error

kyokukou:


bdc34 (changes requested):
Please add a unit test with one or two of the URLs from the jira ticket and ensure the result is not a 500.

bdc34 (approved):

PR #519, opened 2024-03-06
[ArxivCE-1316] subscribe to submission published pubsub and push files

ntai-arxiv:


script/sync_prod_to_gcp/Makefile
```diff
@@ -29,5 +33,15 @@ test/arxiv-development_arxiv-stanza.json:
...
+sync-to-gcp-queue:
+	-gcloud pubsub topics create ${SYNC_TO_GCP_TOPIC} --message-retention-duration=31d --message-encoding=json --project ${ARXIV_PROJECT_ID} --schema p2p
+
+sync-to-gcp-sub: 
+	gcloud pubsub subscriptions create ${SYNC_TO_GCP_SUB_ID} --topic=${SYNC_TO_GCP_TOPIC} --ack-deadline=10 --min-retry-delay=60 --max-retry-delay=600 --max-delivery-attempts=100
+
+install-sync-to-gcp-service:
+	sudo -H install -m 0644 resource/systemd/sync-to-gcp@.service /etc/systemd/system/
+	sudo systemctl daemon-reload
+	@echo "Please run sudo systemctl enable sync-to-gcp@sync-node.service"
```
norbusan, line 47: Should we add `--now` to the `systemctl` call to ensure it is also started, not only enabled, otherwise one needs a reboot to activate it?

ntai-arxiv: You are right. However, this is one time only on one host CIT sync-node, and it is more or less a reminder to myself when I deploy this.
When it is started, the service starts consuming the queue, and did not want to start without monitoring the logging. 
I know I can start it with systemctl start. But, I'm pretty sure I want to start it as e-prints and by hand at first.

norbusan: Ok, thanks!

norbusan (approved):
LGTM

script/sync_prod_to_gcp/README.md
```diff
@@ -99,5 +99,25 @@ OTOH, because of this, it is rather important for Stanza to be running.
...
+
+A systemd service "submssions-to-gcp" subscribes to GCP pub/sub queue for published and 
+copy the submission tarball and abstract file to GCP bucket.
+
+
+## Service setup
+
+There is a (template) unit file. The service process runs one process per web node.
+
+    sudo systemctl enable submissions-to-gcp@sync-node.service
```
bdc34, line 114: I usually think of the @ as for systemd with multiple copies of a service. Why are there two copies? What is the difference between dync-node and sync-node? Please add that to this doc.

script/sync_prod_to_gcp/subscribe_submissions.py
```diff
@@ -0,0 +1,174 @@
+import argparse
```
bdc34, line 1: Please add a docstr. What is this? where is it used? what is its purpose?

script/sync_prod_to_gcp/README.md
norbusan: Good point, the service file should not have an `@` in the filename, since we don't reuse this information as in `%i` somewhere in the systemd unit.

script/sync_prod_to_gcp/test/ntai.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/bash
```
bdc34, line 1: Please don't check in a file named "ntai.sh" Please give it a more descriptive name.  Or don't check it in since there is script/sync_prod_to_gcp/submissions-to-gcp.sh ?

bdc34 (changes requested):
Some minor doc changes requested.

bdc34:
Other then the docs and the @ systemd question this looks good. I'm looking forward to having this in place.

script/sync_prod_to_gcp/test/ntai.sh
ntai-arxiv: Done.

ntai-arxiv: It was to use the arxiv-development queue and bucket. Renamed it accordingly.

script/sync_prod_to_gcp/subscribe_submissions.py
```diff
@@ -0,0 +1,208 @@
...
+        return ("",[])
+
+    paper_id = data.get('paper_id')
+    version = data.get('version')
+    xid = Identifier(f"{paper_id}v{version}" if version else paper_id)
+    archive = ('arxiv' if not xid.is_old_id else xid.archive)
+    logger.info("Processing %s", xid.ids)
+    pairs = []
+
+    if xid.has_version:
```
bdc34, line 74: This appears to be the code to figure out what files to move for .abs and src. The .abs file should always be named consistently.

The source may not have a consistent .tar.gz name. (would be nice to have a arxiv-base Metadata.get_src_files() or something) 

Some other possibilities:

{paperidv}.pdf    pdf only submission
{paperidv}.gz       single file submisison
{paperidv}.html.gz    html source submission

I think there is one case in the system where there is a submission with two source files that the admins manually crafted. It would have: {paperidv}.pdf  and {paperidv}.tar.gz 

This was to get a pdf only submission with ancillary files.

ntai-arxiv: I was not intending this to be do-all. But, I guess it's not too far to do-all.

bdc34: If it is new/cross/jref just /data/ftp/arxiv/papers/{YYMM}/{paperidv}.* needs to be synced.

If it is a replacement or wdr additionally /data/orig/arxiv/papers/{YYMM}/{paperid}{version-1}.* needs to be synced.

script/sync_prod_to_gcp/subscribe_submissions.py
```diff
@@ -0,0 +1,213 @@
...
+        json_formatter = ArxivSyncJsonFormatter(**LOG_FORMAT_KWARGS)
+        json_formatter.converter = gmtime
+        json_logHandler.setFormatter(json_formatter)
+        json_logHandler.setLevel(logging.DEBUG if args.debug else logging.INFO)
+        logger.addHandler(json_logHandler)
+        pass
+
+    # Ensure the environment is authenticated
+    if "GOOGLE_APPLICATION_CREDENTIALS" not in os.environ:
+        logger.error("Environment variable 'GOOGLE_APPLICATION_CREDENTIALS' is not set.")
```
norbusan, line 205: Do we really want to error out here?
In particular, when running in dev mode, we could rely on the default login credentials from `gcloud login`, which would not require downloading credentials, or copying somewhere very specific.
Every user with access rights could run it immediately.

script/sync_prod_to_gcp/test/dev-suscribe-submissions.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/bash
+# Runs subscribe_submissions.py for arxiv-development.
+# You'd need a service account that can ues both the pub/sub and bucket I/O.
+. sync.venv/bin/activate
+export GOOGLE_APPLICATION_CREDENTIALS=~/sync-test.json
```
norbusan, line 5: See above, maybe better rely on `gcloud login` for local development?

script/sync_prod_to_gcp/subscribe_submissions.py
ntai-arxiv: The script is intended to run on CIT sync-node. Only way to get there is to use the VPN. Since other cron jobs do the same, singling this out doesn't make sense.

script/sync_prod_to_gcp/subscribe_submissions.py
ntai-arxiv: @bdc34 I think I did what mentioned here. Please review again.

mnazzaro (approved):

script/sync_prod_to_gcp/subscribe_submissions.py
```diff
@@ -0,0 +1,279 @@
...
+    # I think there is one case in the system where there is a submission with two source files
+    # that the admins manually crafted. It would have: {paperidv}.pdf and {paperidv}.tar.gz
+    # This was to get a pdf only submission with ancillary files.
+
+    # Now, the submission file extension is provided in the message
+    src_ext = data.get('src_ext')
+    # Not sure I have to do this but better safe than sorry
+    src_ext = src_ext.strip() if src_ext else None
+    if src_ext:
+        submission_exts = [src_ext.strip(), ".abs"]
```
bdc34, line 94: Minor: redundant to strip() on line 92

script/sync_prod_to_gcp/subscribe_submissions.py
```diff
@@ -0,0 +1,279 @@
...
+    else:
+        src_ext = None
+        # cast a wider net for older queue element
```
bdc34, line 97: Comment might be incorrect, .pdf .gz .html.gs are possible in new submissions

script/sync_prod_to_gcp/subscribe_submissions.py
```diff
@@ -0,0 +1,279 @@
...
+    # if it is new/cross/jref just /data/ftp/arxiv/papers/{YYMM}/{paperidv}.* needs to be synced.
+    # If it is a replacement or wdr additionally /data/orig/arxiv/papers/{YYMM}/{paperid}{version-1}.* needs to be synced.
+    if publish_type in ["rep", "wdr"]:
+        prev_version = 1
+        try:
+            prev_version = max(1, int(version) - 1)
+        except:
+            pass
+        versioned_parent = f"{ORIG_PREFIX}{archive}/papers/{xid_latest.yymm}"
+        for dotext in submission_exts:
```
bdc34, line 123: I'm not totally sure submission_exts is going to be the full list here. If a src_ext is sent in a wdr message then the submission_exts list isn't going to be correct for the version-1 copy?

bdc34 (changes requested):
One question in subscribe_submissions.py about the submission_exts list.

The other two questions if that file are minor.

bdc34 (approved):

PR #522, opened 2024-03-11
Change how download formats are handled to avoid reading from GS 

bdc34:
/abs was building the possible formats for a paper by doing a list of a key prefix at GS. This is a "class A" operation and while inexpensive, doing it a half a million times adds up. 

This PR uses just the source_format and source_flags from the arXiv_metadata table to figure out the download format options for a paper.


browse/commands/check_paper_formats.py
```diff
@@ -0,0 +1,61 @@
...
+bp = Blueprint("check", __name__)
+
+@bp.cli.command("paper_formats")
+@click.argument("yymm")
+def check_paper_formats(yymm: str) -> None:
+    """Checks formats for yymm."""
+    query = (db.session.query(Metadata.paper_id, Metadata.version,
+                            Metadata.source_format,
+                            Metadata.source_flags, Metadata.source_size)
+                .filter(or_(Metadata.paper_id.like(f"%/{yymm}%"),
```
kyokukou, line 25: What are you trying to filter out here?

browse/services/documents/format_codes.py
```diff
@@ -69,67 +68,66 @@ def formats_from_source_flag(source_flag: str,
     F - PDF only
         PDF-only submission with .tar.gz package (likely because of anc files)
     """
-    formats = []
-    if not source_flag:
-        source_flag = ''
-    if not format_pref:
-        format_pref = ''
+    if isinstance(source_flag, SourceFlag):
+        source_flag = source_flag.code
```
jonathanhyoung, line 72: I wasn't going to comment on Union[str, SourceFlag], but this is pretty ugly. Note that the type of source_flag changes here to : str !

browse/commands/check_paper_formats.py
```diff
@@ -0,0 +1,61 @@
...
+                            Metadata.paper_id.like(f"{yymm}.%"))))
+
+    rows = query.all()
+    if not rows:
+        print(f"No rows found.")
+    a_store: ArticleStore = get_article_store()
+    results = {}
+    for row in rows:
+        paper_id, version, src_fmt, source_flags, source_size = row[0], row[1], row[2], row[3], row[4]
+        print(f"Doing {paper_id}v{version}")
```
kyokukou, line 35: do you still want print statements?

jonathanhyoung (approved):

browse/commands/check_paper_formats.py
bdc34: Yes, this package is a CLI for gathering some data about the different ways of making the format list for a paper. It is not intended to be used in a flask web app.

browse/services/documents/format_codes.py
bdc34: Ya It would be nice to do this differently. Maybe get rid of the class `SourceFlag` and have the properties on `VersionEntry`?

bdc34: The `SourceFlag` is just a string with some properties to decode its meanings.

browse/services/documents/db_implementation/db_abs.py
```diff
@@ -96,7 +96,8 @@ def _to_docmeta(all_versions: List[Metadata], latest: Metadata, ver_of_interest:
                              source_flag=SourceFlag(ver.source_flags),
                              source_format=ver.source_format,
                              is_withdrawn=ver.is_withdrawn or ver.source_format == "withdrawn"
-                                          or ver.source_size == 0)
+                                          or ver.source_size == 0,
+                             is_current= ver.version == len(all_versions))
```
kyokukou, line 100: why not use the is_current value Metadata already has?

PR #526, opened 2024-03-12
ARXIVCE-1333: pi day donate banner

bmaltzan:
This is the pi-day donation popup banner, and a message in the header to the right of the cornell logo.


bmaltzan:
Screen Shot 2024-03-12 at 3 04 34 PM

bmaltzan:
Screen Shot 2024-03-12 at 3 04 17 PM

jonathanhyoung (approved):

katboboris:
For the message in the header with the Cornell Logo, would it be possible to make it a yellow background with black lettering, and include the smiley bones graphic on the left? Right now I think it is hard to read / does not catch attention. 
Giving Message in Banner Example

Also can you change the end date for the message in header to 3/14/2024 at 11:59PM? I think it will help to have it run an extra day. Thanks!

katboboris (changes requested):
For the message in the header with the Cornell Logo, would it be possible to make it a yellow background with black lettering, and include the smiley bones graphic on the left? Right now I think it is hard to read / does not catch attention. I have attached an example.

Also can you change the end date for the message in header to 3/14/2024 at 11:59PM? I think it will help to have it run an extra day so people will still see it when they click out of the pop up. Thanks!

Stephanie also requested that the pop up banner be styled more closely to the banner provided in the example (cream pop up banner at bottom of screen) I believe because it is a little larger, and we're aiming for "briefly annoying yet effective" on Pi Day. This change request is purely stylistic, so if it's not possible to change in the time given, that's okay.

Giving Message in Banner Example

Giving Week 2020 Banners

bmaltzan:
@katboboris 
- The dates look right to me. We normally use midnight, rather than 11:59, as it's easier to read. But if I'm missing something, let me know what line you are looking at.
- I've updated the text, but you'll need to talk with @SBBCornell if you want to redesign the templates. I think we're out of time though, as we really need to take the time to test multiple devices, etc.

katboboris:
> * The dates look right to me. We normally use 2024-03-15, rather than 2024-03-14 11:59, as it's easier to read. But if I'm missing something, let me know what line you are looking at.

@bmaltzan sorry, I think I originally requested that the message in the header end run a different amount of time from the banner, but 3/14 - 3/15 for both is fine. Thanks for clearing that up!

> * I've updated the text, but you'll need to talk with @SBBCornell if you want to redesign the templates. I think we're out of time though, as we really need to take the time to test multiple devices, etc.

@bmaltzan okay, no worries. it's my fault for requesting so late, so style changes canceled! I'll approve changes. :)

katboboris (approved):

PR #527, opened 2024-03-13
dev->main: Pi-day banner

bmaltzan:
This is in dev, and can be previewed here: https://browse.dev.arxiv.org/

The banner inside the header will show up now.
The second and larger popup banner will activate at midnight.


browse/templates/base.html
```diff
@@ -35,6 +35,16 @@
           <a href="https://www.cornell.edu/"><img src="{{ url_for('static', filename='images/icons/cu/cornell-reduced-white-SMALL.svg') }}" alt="Cornell University" /></a>
         </div>
 
+        {%- if rd_int >= 202403120100 and rd_int <= 202403150100 -%}
+        <div class="column">
+          <p>Pi Day is Giving Day: arXiv depends on donations to operate and keep science open for all.
+              <a href="https://givingday.cornell.edu/campaigns/arxiv" target="_blank">
+              <span style="color:white;text-decoration-line:underline;">Give back to arXiv on 3.14.24!</span>
```
cbf66, line 42: What's 3.14.24?? I thought Pi was 3.14?

(This would worked better in 2015).

katboboris: Lots of things worked better in 2015! (I'm not opposed to dropping the .24 and changing it to 3.14 for the math of it all)

PR #539, opened 2024-03-21
Increases cache-control max age for files

bdc34:
`Requests` for versioned files get the max-age set to 7 days and unversioned requests for one day.

### DIFF
--- a/browse/controllers/files/__init__.py
+++ b/browse/controllers/files/__init__.py
@@ -10,8 +10,27 @@ from arxiv.files import FileObj
 
 
 BUFFER_SIZE = 1024 * 4
+
 CACHE_AGE_SEC_VERSIONED = 60 * 60 * 24 * 7
+"""cache-control max-age for versioned requests.
+
+An example of a versioned request: /pdf/0202.12345v1
+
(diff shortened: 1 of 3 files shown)


browse/controllers/files/__init__.py
```diff
@@ -10,6 +10,8 @@
 
 
 BUFFER_SIZE = 1024 * 4
+CACHE_AGE_SEC_VERSIONED = 60 * 60 * 24 * 7
+CACHE_AGE_SEC_UNVERSIONED = 60 * 60 * 24
```
kyokukou, line 14: not sure we are too worried about it, but this does mean replacements will consistently not be visible at publish/announce time if someone looked at the non versioned paper the day before.

bmaltzan: > not sure we are too worried about it, but this does mean replacements will consistently not be visible at publish/announce time if someone looked at the non versioned paper the day before.

It would, so we also purge the previous nights announcements at a few points.
And yes, it would be terrible for us to announce a replacement, and then send users an older pdf!

bdc34:
I've added comments about the cache control constants.

kyokukou (approved):

PR #540, opened 2024-03-22
Fixes bug with conference proceedings

bdc34:
The html_post_process fn subs in ABS and LIST in the html text for actual looked up abs metadata at time of request.

This was broken due to a recent change that moved it to after the app code so the fn was not in flask's request context. 

Calls `html_post_process` wrapped in `flask.stream_with_context` so it can access the metadata service even after the app code has returned.

Adds a test that should ensure that the `html_post_process` is not having errors due to this flask context problem

https://arxiv-org.atlassian.net/browse/AH-93375

### DIFF
--- a/browse/controllers/files/dissemination.py
+++ b/browse/controllers/files/dissemination.py
@@ -24,7 +24,7 @@ from browse.services.dissemination import get_article_store
 from browse.services.dissemination.article_store import (
     Acceptable_Format_Requests, CannotBuildPdf, Deleted)
 
-from flask import Response, abort, make_response, render_template, request, current_app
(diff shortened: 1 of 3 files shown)


jonathanhyoung (approved):

ntai-arxiv (approved):
https://github.com/arXiv/arxiv-browse/pull/540/files#diff-f24c4764ac94b3bc195e4b2410916950ac031d530f0fb39df3fc548675ec7f5bR72

A bit weird to me but Brian says it is for typing.

cbf66 (approved):

PR #557, opened 2024-04-05
Adds txt_convert to subscribe_submissions.py

bdc34:
Adds ability to listen to txt_convert schema messages, make a payload of files to copy and then copies them.

I plan to run this as a separate process from the publish sync job.


bdc34:
Marked as "Will not do" in jira.

PR #564, opened 2024-04-15
simplify fetching the return format

kyokukou:


bdc34 (approved):
This seems like a nice reduction in complexity.

PR #577, opened 2024-04-22
Temp 2 digit year listings

kyokukou:
Potential pause to the 4 digit year url change until we are confident GCP listings will be kept up. There was a convenient place to bandaid the /list path but not the /year path so the plan would be just to deploy /list until we are confident it is good to go then we can revert this change and also deploy /year

Changed here:
-Band-Aid time period processing function that expects 2 digit years and never redirects
-use 2 digit urls to run listing tests
-skip the specific redirect test
-disable the url change alert

### DIFF
--- a/browse/controllers/list_page/__init__.py
+++ b/browse/controllers/list_page/__init__.py
@@ -129,7 +129,6 @@ def get_listing(subject_or_category: str,
     show
        Number of articles to show
     """
-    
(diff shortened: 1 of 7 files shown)


cbf66 (approved):
I'm approving this in principle, but I did not do a careful review of the code.

Hopefully we'll be able to just back this all out a few days from now.

bdc34 (approved):

PR #582, opened 2024-04-23
added spaces

kyokukou:


cbf66 (approved):
Trying to see if the spaces are right in the template hurts my eyes. Deploy it and we'll see.

PR #602, opened 2024-05-08
Arxivce 1715 add surrogate key header to list

kyokukou:
fastly can use surrogate keys to purge specific sections of the cache

### DIFF
--- a/browse/controllers/__init__.py
+++ b/browse/controllers/__init__.py
@@ -3,8 +3,7 @@
 Each controller corresponds to a distinct browse feature with its own
 request handling logic.
 """
-from datetime import timezone, datetime, timedelta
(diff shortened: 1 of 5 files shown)


browse/controllers/__init__.py
```diff
@@ -51,3 +50,9 @@ def biz_tz() -> ZoneInfo:
         return _arxiv_biz_tz
     else:
         return _arxiv_biz_tz
+
```
bdc34, line 53: Minor: what happens if the old_keys has a key that is in the new keys?

bdc34 (approved):

browse/controllers/__init__.py
kyokukou: fastly ignores duplicates, but i can fix that

PR #605, opened 2024-05-10
Rewrite fs_abs to use ObjectStore so it works with GCS Buckets again

mnazzaro:


browse/services/dissemination/__init__.py
```diff
@@ -1,37 +1,22 @@
...
 
 
-def get_global_object_store(path: str, global_name: str) -> ObjectStore:
-    """Creates an object store from given path."""
-    store = globals().get(global_name)
-    if store is None:
-        uri = urlparse(path)
-        if uri.scheme == "gs":
-            gs_client = storage.Client()
-            store = GsObjectStore(gs_client.bucket(uri.netloc))
```
bdc34, line 28: I didn't notice this before, but I don't love that get_global_object_store has moved to base since it is sort of parameterized on the bucket. 

Some other app might want a obj store for submission source, or log files or audit trails, and those could live in different buckets.

Probably don't change this now but something to keep in mind.

browse/services/documents/__init__.py
```diff
@@ -16,8 +22,9 @@ def get_doc_service() -> DocMetadataService:
 def fs_docs(config: dict, _: Any) -> DocMetadataService:
     """Factory function for file system abstract service."""
     from browse.services.documents.fs_implementation.fs_abs import FsDocMetadataService
-    return FsDocMetadataService(config["DOCUMENT_LATEST_VERSIONS_PATH"],
-                                config["DOCUMENT_ORIGINAL_VERSIONS_PATH"])
+    return FsDocMetadataService(
+        get_global_object_store(config["ABS_PATH_ROOT"], '_doc_latest_versions_store'),
```
bdc34, line 26: Oh, never mind the above comment, this seems fine.

bdc34 (approved):

PR #607, opened 2024-05-13
Arxivce 1734 fix plain archive path

kyokukou:


jonathanhyoung (approved):

browse/templates/archive/archive_list_all.html
```diff
@@ -10,7 +10,7 @@ <h1 class="header-breadcrumbs"><a href="/"><img src="{{ url_for('static', filena
 {%- block content %}
   <h1>Archives</h1>
 
-  {% if bad_archive is not none %}
+  {% if bad_archive is not none and bad_archive != "list" %}
```
bmaltzan, line 13: Should this be "None"?

kyokukou: jinja considers both the same thing

PR #611, opened 2024-05-14
A lot of changes

mnazzaro:
- Light mode as default for HTML conversions
- FS Abs works again using obj store
- Cloud run stack trace fixes from base
- Fixes for /archive page
- Surrogate keys for selective fastly cache clearing


browse/static/js/feedbackOverlay.js
```diff
@@ -676,4 +676,7 @@ document.addEventListener("DOMContentLoaded", () => {
     });
 
     document.getElementById('myFormContent').onsubmit = submitBugReport;
+
```
cbf66, line 679: If this is going to override the user's browser default, I don't think we should do this.

mnazzaro: Why not? This is convenient for the user because it allows us to prefill the github issue template for them. It also lets us put the data into the feedback table at the same time

jonathanhyoung (approved):

bdc34 (approved):

PR #626, opened 2024-05-30
Handle PDF ids in reasons.json file

bdc34:
PDF requests for ids in the `gs://arxiv-production-data/reasons.json` file get a 404 and a special page. 

This mimics what happens in legacy.

https://arxiv-org.atlassian.net/browse/ARXIVCE-1783


bdc34:
I added a test for genpdf.

PR #627, opened 2024-05-31
Arxivce 1775 prevent encrypted source download

kyokukou:


browse/controllers/abs_page.py
```diff
@@ -147,6 +147,9 @@ def get_abs_page(arxiv_id: str) -> Response:
                     response_data["higher_version_withdrawn_submitter"] = _get_submitter(abs_meta.arxiv_identifier,
                                                                                          ver.version)
 
+        response_data["encrypted"] = abs_meta.version_history[abs_meta.version - 1].source_flag.source_encrypted
```
bdc34, line 150: Could this be `abs_meta.get_requested_version().source_flag.source_encrypted`?

browse/controllers/files/__init__.py
```diff
@@ -102,6 +102,17 @@ def withdrawn(arxiv_id: Identifier, had_specific_version: bool=False) -> Respons
...
 
+def not_public(arxiv_id: Identifier, had_specific_version: bool=False) -> Response:
+    """ Returned for pages whose source is encrypted/ made not public by the author.
+    Sets expire to one year, max allowed by RFC 2616"""
+    if had_specific_version:
+        headers = {'Cache-Control': 'max-age=31536000'}
+    else:
+        headers = {'Cache-Control': maxage(False)}
+    return make_response(
+        render_template("dissemination/not_public.html",arxiv_id=arxiv_id), 403, headers
```
bdc34, line 113: I'm tempted to do a status other than 403 to avoid ambiguity with crowdsec and real unauth .

bdc34 (commented):
This will still allow download of the encrypted source from `/src` is that fine?

browse/controllers/files/__init__.py
kyokukou: this was the behavior the legacy system had before but im happy to change it if it causes problems down the line

browse/templates/abs/extra_services.html
```diff
@@ -54,7 +54,7 @@
       {%- if format == 'latexml' -%}
   <li><a href="{{ latexml_url }}" class="abs-button" id="latexml-download-link">HTML (experimental)</a></li>
       {%- endif -%}
-      {%- if format == 'src' and current-%}
+      {%- if format == 'src' and current and not encrypted-%}
```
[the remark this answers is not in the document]
kyokukou: no idea

kyokukou:
> This will still allow download of the encrypted source from `/src` is that fine?

what makes you say this? the /src path now returns the html template

browse/controllers/files/__init__.py
jweiskoff: How about `405`, or `451` instead? https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors

tests/dissemination/test_encrypted.py
```diff
@@ -0,0 +1,13 @@
+
+def test_encrypted(client_with_test_fs):
+    """Tests when there is no dissemination file for an existing tex version"""
+    resp = client_with_test_fs.get("/src/0704.0380")
```
bdc34, line 4: :+1:

browse/controllers/files/__init__.py
bdc34: 403 is fine.

bdc34 (approved):

browse/controllers/files/__init__.py
kyokukou: 405 isnt correct. 451 seems...extreme... and possibly slightly inaccurate.
i could see 200 being applicable because we returned the template OK or 403 because thats not a page they are allowed to access, there should be no links to the page on the site

jweiskoff: although maybe not `451`, as there's an IIS code with the same id, and it looks like it's handled differently in some MS browsers. https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#Internet_Information_Services

browse/controllers/abs_page.py
kyokukou: oh theres a function for this! Thats way more convenient :)

browse/controllers/files/__init__.py
jweiskoff: Another alternative: `204`. Since there's no real error on the client side, we're just not returning content for the request.

browse/routes/dissemination.py
```diff
@@ -86,6 +86,7 @@ def format(arxiv_id: str, archive: Optional[str] = None) -> Response:
     data = {"arxiv_id": arxiv_identifier.id,
             "arxiv_idv": arxiv_identifier.idv,
             "abs_meta": abs_meta}
+    data["encrypted"] = abs_meta.version_history[abs_meta.version - 1].source_flag.source_encrypted
```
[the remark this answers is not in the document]
jweiskoff: No, it should not allow download of the encrypted source.

browse/templates/abs/extra_services.html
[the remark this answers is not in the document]
jweiskoff: That also seems wrong. All versions work from the url construction, so we should/could be showing it across the board.

jweiskoff (approved):

PR #637, opened 2024-06-12
html and pdf paths still available if source is encrypted

kyokukou:


browse/services/dissemination/article_store.py
```diff
@@ -224,6 +224,7 @@ def dissemination(self,
 
         `path` is additional data for html and anc requests.
         """
+        print(format)
```
bmaltzan, line 227: Keeping in?

bmaltzan (approved):

PR #647, opened 2024-06-26
added missing comma, fixes DOI formatting

jweiskoff:
There's a comma missing, which causes the bibtex output to not validate correctly.


collares:
Great, thanks! I think the DOI field is also missing a comma before the `\n`, if I understand correctly.

kyokukou (approved):

PR #653, opened 2024-07-05
Fixes broken call to closure _get_files_for_html()

bdc34:
ARIXVCE-1928


bdc34:
Chatted with Tai about this. He is working on other stuff in these files and doesn't need this. The code works without these fixes changes.

PR #676, opened 2024-07-24
Graceful latexml DB down, part 1

bdc34:
Adds a `ignore_errors` arg to `db_handle_error` decorator that we wrap many db calls with to have uniform error handling.

Sets `ignore_errors` to `True` on latexml DB calls. This should reduce cases where latexml DB down causes the service to error out.

This PR can be deployed immediately. An additional change in arxiv-base to handle the rollback is still needed to complete the graceful latexml DB down feature. It would be ideal if we could set arxiv-browse so it does not try to roll anything back since it is read only.

ARXIVCE-2433

### DIFF
--- a/browse/services/database/__init__.py
+++ b/browse/services/database/__init__.py
@@ -56,8 +56,16 @@ app_config = get_application_config()
 tz = gettz(app_config.get("ARXIV_BUSINESS_TZ"))
 
 
-def db_handle_error(db_logger: Logger, default_return_val: Any) -> Any:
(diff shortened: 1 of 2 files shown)


cbf66 (commented):
Can we make it all READUNCOMMITTED while we're at it?

DavidLFielding (approved):
Looks good.

bdc34:
> Can we make it all READUNCOMMITTED while we're at it?

Let's do one thing at a time.

PR #677, opened 2024-07-24
Graceful latexml DB down, part 1 Deploy

bdc34:
Deploy of PR https://github.com/arXiv/arxiv-browse/pull/676

Text from that PR:

Adds a ignore_errors arg to db_handle_error decorator that we wrap many db calls with to have uniform error handling.

Sets ignore_errors to True on latexml DB calls. This should reduce cases where latexml DB down causes the service to error out.

This PR can be deployed immediately. An additional change in arxiv-base to handle the rollback is still needed to complete the graceful latexml DB down feature. It would be ideal if we could set arxiv-browse so it does not try to roll anything back since it is read only.

ARXIVCE-2433


jweiskoff (approved):

ntai-arxiv (approved):
LGTM

PR #684, opened 2024-08-05
[ARXIVCE-2462] When the tex compiation request is stuck for 30 minutes, send email to the support team. NOTE - the systemd service is already running on the sync node. 

ntai-arxiv:


script/sync_prod_to_gcp/webnode_pdf_request.py
```diff
@@ -125,6 +128,27 @@ def ping_callback(message: Message) -> None:
             message.ack()
             return
 
+        message_age: timedelta = datetime.utcnow() - message.publish_time
+        compilation_timeout = int(os.environ.get("TEX_COMPILATION_TIMEOUT_MINUTES", "30"))
+        if message_age > timedelta(minutes=compilation_timeout):
+            help_needed = os.environ.get("TEX_COMPILATION_RECIPIENT", "help@arxiv.org")
+            subject = f"TeX compilation for {paper_id}v{version} failed"
+            mail_body = f"Hello EUTA,\nTex compilation for {paper_id}v{version} has failed. Please resolve the issue.\n\nThis message is generated by a bot on arxiv-sync.serverfarm.cornell.edu.\n"
```
jweiskoff, line 136: EUST not EUTA.

This isn't very helpful because it doesn't tell us which node it's failing on, nor does it provide the link to review the tex log. This is better than nothing, but is it also killing the long-running tex process? we shouldn't allow them to run that long, I figure.

jweiskoff (changes requested):

script/sync_prod_to_gcp/webnode_pdf_request.py
```diff
@@ -125,6 +128,27 @@ def ping_callback(message: Message) -> None:
...
+            cmd = ["/usr/bin/mail", "-r", "developers@arxiv.org", "-s", subject, help_needed]
```
norbusan, line 137: Will that work in the gcloud in the future?

ntai-arxiv: This is for a systemd service running on sync-node at CIT so this is tailored for it. The resulting PDF is checked on the CIT's file server so it's not possible to run on GCP.

ntai-arxiv: More over, once genpdf becomes the solo compiler, this will go away.

jweiskoff (approved):
I still think it would be better to know which web node it fails on rather than just a generic message saying it failed. Although Charles wanted to discuss whether this should be going to us first. 
Let's do that before merging.

ntai-arxiv:
> I still think it would be better to know which web node it fails on rather than just a generic message saying it failed. Although Charles wanted to discuss whether this should be going to us first. Let's do that before merging.

Each attempt of ping (pdf request) goes to different web node (web5 - web9), so after 30 minutes, every web node is tried and failed after 30 minutes. 

https://console.cloud.google.com/logs/query;query=labels.log_type%3D%22arxiv_sync2gcp_log%22%0ASEARCH%2528%22nack%20message%22%2529;summaryFields=:false:32:beginning;cursorTimestamp=2024-08-05T14:02:10.120124999Z;duration=PT1H?referrer=search&project=arxiv-production

In the each log entry, there is a field "count" which varies from 0-4 which corresponds to web5 to web9.

PR #701, opened 2024-08-17
[ARXIVCE-1756] Grand unified sync-to-gcp

ntai-arxiv:
This is intended to replace 2 services - one is syncing the files, the other is to ask webnode the PDF.

- Reads the paper IDs, etc. from pubsub (let's call each queue element / paper ID "job")
- Figures out the files that should be on the GCP
- Copies the files that's not there
- For /ftp directory, the extra files are moved to "/trash"
- Retries based on the accept/reject of the job completion.
- Relies on the queue back off for retries
- When the job is stuck for long time (30 minutes), send a email to admin team and slack for us.  This means that - rather than the queue element is stuck in the queue, the email / admin ticket is the queue.

- It starts the local http server to mimic the web node and it copies the pdf to the correct location. It can also 404 or timeout
- Unit-testing makes sure the correct files are listed from the job request.
- It can trash the obsolete (GCP) files - which is the point of this ticket.

image

I need to deploy this and then monitor the logs to see the existing cron jobs of syncing does nothing.


script/sync_prod_to_gcp/submissions_to_gcp.py
```diff
@@ -0,0 +1,928 @@
+"""
+submissions_to_gcp.py is an app that gets the published arxiv ID from the pub/sub queue on GCP,
+check the files exist on CIT and uploads them to GCP bucket.
+
```
bdc34, line 4: Please add something like:

"For each announced paper ID it will:
sync the /ftp files and ensure only current version files exist 
sync the abs and source files to /orig if needed
If TeX source it will build the PDF on a CIT web node and sync those to ps_cache
If HTML source it will build the HTML and sync those to the ps_cache"

script/sync_prod_to_gcp/submissions_to_gcp.py
```diff
@@ -0,0 +1,928 @@
...
+    for entry in desired_state:
+        local = entry["cit"]
+        remote = entry["gcp"]
+        entry_type = entry['type']
+
+        if entry_type == "html-cache":
+            # "html-cache" is needed on CIT as it is the root dir of HTML submission, but
+            # it is NOT a bucket object, and nothing to upload. Subsequent object copy becomes
+            # effectively creating the "directory" on GCP.
+            logger.debug("Skipping [%s]: %s -> %s", entry_type, local, remote, extra=log_extra)
```
bdc34, line 804: I'm concerned that html source papers are not getting synced. For each html paper what is needed:
1. build the html at a CIT web node by requesting the html
2.sync the html from ps_cache to GCP

bdc34 (commented):
The README.md needs a section before the "History" that describes what the current state of the sync is, what it does, etc. 

The 3 of scripts under the sync_prod_to_gcp are a little confusing. There is a systemd config just for submissions_to_gcp so it seems that this is the only one in use now. Could the sync_published ones be moved to their own dir? Could all the other scripts be moved to their own dirs?

Does this preprocess the html sourced papers and then sync those to arxiv-production-data:/ps_cache? This is a requirement. 
Please do not move objects to /trash on the arxiv-production-data. It will build up  cruft that is useless.  It has no analog on the CIT SFS. It will not be clear what process caused it be created.

script/sync_prod_to_gcp/submissions_to_gcp.py
ntai-arxiv: I decided to remove the condition of checking to see the HTML root dir exists. I think that was a bug that if someone else asks the HTML first (or say before the submission gets to here), it may not have populated the html files. With the change, it should scoop up the files regardless of non-existance of HTML root dir, and tries to send the files. So thanks for making me think this.

ntai-arxiv:
> The 3 of scripts under the sync_prod_to_gcp are a little confusing. There is a systemd config just for submissions_to_gcp so it seems that this is the only one in use now. Could the sync_published ones be moved to their own dir? Could all the other scripts be moved to their own dirs?

submissions_to_gcp.py uses sync_published_to_gcp.py's ensure_FOO. And, for how, the cron job is still using sync_published_to_gcp.py/sync_published.sh. If any, webnode_pdf_request.py is def. obsolete. 

I will think about how to consolidate the 3 services. However, I'd like to keep the files as is since the deployment relies on these python and shell scripts.

bdc34 (approved):

script/sync_prod_to_gcp/submissions_to_gcp.py
```diff
@@ -345,58 +451,61 @@ def maybe_update_metadata(self):
...
+        except Exception as exc:
+            msg = f"Error with {abs_path}"
+            logger.warning(msg, exc_info=True)
+            raise BrokenSubmission(msg) from exc
+
+        if dates:
+            self.set_published_versions(dates)
+
+            published = self.get_version_metadata(self.version)
+            # The published abs is the source of truce.
```
bdc34, line 479: Typo: truce -> truth 
(unless this is intended as a pun?)

bdc34 (approved):

PR #702, opened 2024-08-19
Changes to Dockerfile

bdc34:
Turns off FLASK_DEBUG

Removes apt update for repeatability and caching

Consolidates some env vars

Moves gunicorn to pyproject.toml for depbot checking

Skips install of dev packages

Skips install of tests


jonathanhyoung (approved):
The e-prints (non-root user) pattern is good - but this PR removes the chown/chmod which I thought was always needed - ?

There appears to be a lot of redundancy between the ENV command at line 12 and the ENV commands at lines 51-60. Can they be combined?

bdc34:
>  this PR removes the chown/chmod which I thought was always needed - ?
> 

The chmod is only needed if the app needs to write to those files. The removed chmod was to a test directory that the tests needed to run. But the tests are no run in this image, it is just for the cloud run app.

bdc34:
> There appears to be a lot of redundancy between the ENV command at line 12 and the ENV commands at lines 51-60. Can they be combined?

I think they are combined as I'd like them to be. I don't want to accidentally bump up the log level of pip or poetry by putting that one early.

PR #709, opened 2024-08-22
accessible links in browse 

bdc34:
Desc from issue:
links in browser do not have sufficient contrast (luminosity) to pass a11y wcag guildelines. 
Users have strong feelings about changes to browse. 
Find the lowest impact solution and implement it with care.

Most recent comment:
Shamsi Brinn 
February 25, 2020 at 2:34 PM
Update: On hold for now (and set to ‘will not do’ status). Will hold off on deploying unless we need to if Cornell begins accessibility site testing. This may be moot by then if we’ve already integrated the larger redo of browse that is our ultimate goal. 

https://arxiv-org.atlassian.net/browse/ARXIVNG-3009


bdc34:
Shamsi said via slack: "#709: I think this one should be tossed. We should address in an overall redesign, and not try and remediate the older arXiv bit by bit."

PR #716, opened 2024-08-26
ARXIVCE-2902 Revise /cookies page

cbf66:
1. Eliminate default download format preference. It offered formats like DVI that we don't even support anymore, and had DPI settings for PDF files that I don't think do anything.
2. Eliminate the ability to choose your downloads to come from mirrors
3. Rewrite the text to simplify it.
4. Add code to use X-Forwarder-For IP address for the institutional banner computation so it will get the proper IP address even if the request came through a CDN.


norbusan (approved):
LGTM, mirrors are gone, format selection

dginev:
I did a cookie audit today, and updated this PR to the best of my understanding of our cookie coverage.

Some details of note:

- We resolved in ARXIVCE-2902 that we can link to cookies set dedicated UI components. I have added links to reach them from `/cookies`.
- We should probably defer a toggle to "disable all banners" until 1) we have a naming convention for their cookies and 2) we are sure we want to enable that as a "preference".
- [OpenURL settings](https://arxiv.org/openurl-cookie) has been deemed inactive, so I won't add a link to it in this PR.

I also separated out the X-Forwarded-For feature in its own PR, at #808 

Reviews welcome.

dginev (approved):

PR #732, opened 2024-09-23
Arxivce 1756 abs parse fix

ntai-arxiv:
Fix for .abs file not parseable as utf-8. 

Turn off the webnode ping for pdf as it is retiring (replaced by submissions_to_gcp). 

A litte more logging for tracking the synced files.


dginev (approved):
✅

bdc34 (approved):

PR #755, opened 2024-10-02
Switch to rival-sans font family + related patches

dginev:
This PR uses a new typekit by @SBBCornell which allows us to:
 - use `rival-sans` instead of `freight-sans-pro`
 - use `rival-sans-narrow` instead of `ibm-plex-sans-condensed`
 - keep using `stix-two-math` for math
 - try out the same font-family for text and headings
 - deprecate reliance on `ar5iv-fonts.css` (which held a second typekit asset)

I also patched a couple of CSS selector issues, which should activate the font-family chosen for the ToC and arXiv header and footer.

PR #768, opened 2024-10-17
update fonts for HTML papers

dginev:
Upstreams #755 from develop.

One ops detail to keep an eye on is that we are adding a self-hosted font file and pointing to it via:
```
/static/browse/0.3.4/fonts/STIXTwoMath-Regular.woff2
```


SBBCornell (approved):
looks good

PR #770, opened 2024-10-17
deploy banner for next week

kyokukou:
for open access week


bdc34 (approved):
This looked good when I tried it.

PR #771, opened 2024-10-18
Arxivce 2683 disable citations for wdr paper

kyokukou:


browse/controllers/abs_page.py
```diff
@@ -384,15 +384,19 @@ def _get_submitter(arxiv_id: Identifier, ver:Optional[int]=None) -> Optional[str
         return None
 
 def _show_refs_cites(arxiv_id: Identifier) -> bool:
-    HIDDEN_IDS=["2307.10651"]
-    if arxiv_id.id in HIDDEN_IDS:
+    NO_REFS_IDS=[
```
jweiskoff, line 387: wouldn't this be better done as a feature flag in the db rather than hard coding a specific paper into the python source?

kyokukou:
resulting appearence 
image

browse/controllers/abs_page.py
bdc34: DB feature flag would take longer to roll out. We are hoping to do something like that in the future.

bdc34 (approved):

PR #777, opened 2024-10-22
Cache maxage set to 1 year ARXIVCE-2719

bdc34:
Sets the Surrogate-Control maxage to one year. Surrogate-Control sets only the fastly (or similar) caching.


browse/config.py
```diff
@@ -15,6 +15,7 @@
 DEFAULT_DB = "sqlite:///tests/data/browse.db"
 TESTING_LATEXML_DB = 'sqlite:///tests/data/latexmldb.db'
 
+DAY = 60 * 60 * 24 # one day of seconds
```
kyokukou, line 18: nice :D

kyokukou (approved):

PR #788, opened 2024-11-02
[arXivCE-2763] Submission GCP was lacking the postscript submission (…

ntai-arxiv:
…foo.ps.gz).


script/sync_prod_to_gcp/submissions_to_gcp.py
```diff
@@ -532,6 +532,12 @@ def is_html_submission(self) -> bool:
         assert(self.source_format)
         return self.source_format == "html"
 
+    @property
+    def is_ps_submission(self) -> bool:
+        """is a HTML submission"""
```
norbusan, line 537: That comment should be "is a PostScript submission" or similar

norbusan (approved):
Besides the small issue with the comment, fine for me! Thanks

script/sync_prod_to_gcp/submissions_to_gcp.py
ntai-arxiv: oops.

PR #789, opened 2024-11-08
Add config for reasons file path

bdc34:
Need ability to configure the reasons.json file. 
ARXIVCE-2772

### DIFF
--- a/browse/config.py
+++ b/browse/config.py
@@ -132,6 +132,13 @@ class Settings(arxiv_base.Settings):
     `./testing/data/` for testing data. Must end with a /
     """
 
+    REASONS_FILE_PATH: str = "DEFAULT"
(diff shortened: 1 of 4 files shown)


jweiskoff (approved):

browse/config.py
```diff
@@ -132,6 +132,13 @@ class Settings(arxiv_base.Settings):
     `./testing/data/` for testing data. Must end with a /
     """
 
+    REASONS_FILE_PATH: str = "DEFAULT"
+    """Full path to `reasons.json` file.
```
jweiskoff, line 136: at some point we'll want a page for the EUST to be able to curate this list.

bdc34: Ya, it should go in the DB.

PR #808, opened 2024-12-06
ARXIVCE-2902 use first X-Forwarded-For IP in /institutional_banner

dginev:
SSIA

### DIFF
--- a/browse/routes/ui.py
+++ b/browse/routes/ui.py
@@ -148,7 +148,12 @@ def catchup(subject:str, date:str) -> Response:
 @blueprint.route("institutional_banner", methods=["GET"])
 def institutional_banner() -> Any:
     try:
-        result = get_institution(request.remote_addr)
+        forwarded_ips = request.headers.getlist("X-Forwarded-For")
+        if len(forwarded_ips)>0:
+            ip = forwarded_ips[0]
+        else:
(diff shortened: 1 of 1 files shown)


browse/routes/ui.py
```diff
@@ -148,7 +148,12 @@ def catchup(subject:str, date:str) -> Response:
 @blueprint.route("institutional_banner", methods=["GET"])
 def institutional_banner() -> Any:
     try:
-        result = get_institution(request.remote_addr)
+        forwarded_ips = request.headers.getlist("X-Forwarded-For")
+        if len(forwarded_ips)>0:
+            ip = str(forwarded_ips[0]).split(',')[0]
```
jonathanhyoung, line 153: Can't the forwarded ips be space-delimited, too?

Do you need code to strip the ip string of leading and trailing spaces?

Do you need more robustness / error checking? What if forwarded_ips is None?

dginev: Done. They are defined as [comma-separated](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax), so this should handle the specified syntax.

dginev:
Rebased on develop and ensured type-checking is happy.

cbf66 (approved):

PR #813, opened 2024-12-12
remove close option that is not i understand.

kyokukou:
All must understand


cbf66 (approved):
It is a "close" element. But it's redundant with the "I understand" button, and leaves some ambiguity as to whether the user acknowledged the new privacy policy or simply swatted the notice away.

PR #814, opened 2024-12-12
ARXIVCE-2718 Ensure https scheme when using url_for and _external=True

dginev:
This is one possible solution to avoiding an http-to-http redirect, as reported in ARXIVCE-2718.

The PR explicitly requests that `url_for` construct `https` URLs, when preparing external/absolute links (as contrasted to internal/relative links).


dginev:
While this PR would be one approach to resolve the immediate problem, on second thought it would make the localhost development workflow rather painful. So I retract my suggestion.

PR #821, opened 2025-01-09
removes old download venues from the monthly downloads page

jweiskoff:
Removes mention of the mirrors, as well as removes older info regarding non-web downloads from the stats page.


jimentwood (approved):
Looks good

PR #831, opened 2025-01-28
[ARXIVCE-3081] ensure_pdf to work with old style arxiv ID

ntai-arxiv:
- The fix itself is a single liner change in ensure_pdf().
- The logging around is a little more verbose when it does not get PDF. 
- The test "test_ask_pdf_for_old_arxiv_id" added. 
- The set up removes the PDF in sync-test bucket. 

To run the test, you need a GCP credential so it is not part of git workflow. Once we shift the CI/CD to Drone, we'd be able to run the tests on GCP. For now, run tests by hand. The test result attached.

image


script/sync_prod_to_gcp/sync_published_to_gcp.py
```diff
@@ -390,7 +390,7 @@ def ensure_pdf(session, host, arxiv_id, timeout=0, protocol = "https", source_mt
     timeout = PDF_WAIT_SEC if not timeout else timeout
     archive = ('arxiv' if not arxiv_id.is_old_id else arxiv_id.archive)
     pdf_file = Path(f"{PS_CACHE_PREFIX}/{archive}/pdf/{arxiv_id.yymm}/{arxiv_id.filename}v{arxiv_id.version}.pdf")
-    url = f"{protocol}://{host}/pdf/{arxiv_id.filename}v{arxiv_id.version}.pdf"
+    url = f"{protocol}://{host}/pdf/{arxiv_id.idv}.pdf"
```
bdc34, line 393: :+1:

script/sync_prod_to_gcp/submissions_to_gcp.py
```diff
@@ -773,26 +773,31 @@ def submission_message_to_file_state(data: dict, log_extra: dict, ask_webnode: b
                     WEBNODE_REQUEST_COUNT = (WEBNODE_REQUEST_COUNT + 1) % n_webnodes
                     host, n_para = CONCURRENCY_PER_WEBNODE[min(n_webnodes - 1, max(0, my_tag))]
                     protocol = "http" if host.startswith("localhost:") else "https"
+                    log_extra['webnode'] = host
```
bdc34, line 776: I see that it is difficult to get the URL used to request the PDF due to  ensure_pdf signature. :+1:

bdc34 (approved):

PR #837, opened 2025-03-07
ScienceCast filter update

bdc34:
Per discussion with ScienceCast on 3/6, update the filter on our end to look for submissions with astro-HE as primary, announced since Dec 1 2024, and that are have tex source. 

Minor change from the ticket: it said ", and that are NOT pdf-only. " But I assumed sciencecast cannot handle other strange submissions. Also filtering out HTML, PS and DOCX.


kyokukou (approved):
fancy!

PR #854, opened 2025-04-02
Master merge of PDFLaTeX papers bug fix

bdc34:
Master merge of PDFLaTeX papers should not display postscript download options on the Formats page

#853 from arXiv/develop


bdc34:
Mistake.

PR #864, opened 2025-04-29
fixes modal close for cite.js

jweiskoff:
As reported in https://arxiv-org.atlassian.net/browse/ARXIVCE-3546.

### DIFF
--- a/browse/static/js/cite.js
+++ b/browse/static/js/cite.js
@@ -47,7 +47,7 @@ $(document).ready(function() {
         doi = metadoi ? metadoi.content: ''
 
         if(!setup){
-            $('.bib-modal-close').click(function(){$('#bib-cite-modal').hide()})
+            $('.bib-modal-close').click(function(){$('#bib-modal').hide()})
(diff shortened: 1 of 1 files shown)


browse/static/js/cite.js
```diff
@@ -47,7 +47,7 @@ $(document).ready(function() {
         doi = metadoi ? metadoi.content: ''
 
         if(!setup){
-            $('.bib-modal-close').click(function(){$('#bib-cite-modal').hide()})
+            $('.bib-modal-close').click(function(){$('#bib-modal').hide()})
```
dginev, line 50: That won't work - the ` $('.bib-modal-close')` here hijacks the close button of the `CiteModal` component in `CiteModal.tsx`. If you make it point to the other id, it will stop closing the currently correct id for the "export BibTeX citation" button above.


So either it needs to close both modals, or the cite.js close button needs an id and then the selector on which `.click` gets bound is switched to that id.

jweiskoff:
This is deeper water than I can tread in.

PR #867, opened 2025-05-08
README cleanup and adds ruff

bdc34:


kyokukou (approved):
yay for cleanup! but it looks like your build failed

bdc34:
> yay for cleanup! but it looks like your build failed

I'm trying to get ruff working and it should be simple...

PR #870, opened 2025-05-20
Default timeout of event subsuming is now 12 hours, and the slack notification going out after 30 minutes.

ntai-arxiv:


jweiskoff (approved):

script/sync_prod_to_gcp/submissions_to_gcp.py
```diff
@@ -1043,24 +1075,42 @@ def submission_callback(message: Message) -> None:
...
+                if slacking.returncode != 0:
+                    logger.warning("Failed to send notification: %s", id_v, extra=log_extra)
+            except:
+                logger.warning('Slacking for %s did not work', id_v)
+                pass
+            error_state_file.report()
+
+
     # Deadline exceeded?
     if message_age > timedelta(minutes=compilation_timeout):
```
bdc34, line 1105: This section already seems like it would effectively prevent a flood of failure calls by only sending messages when the message is older than the compile_timeout.

bdc34 (commented):
There is already a system in submissions_to_gcp.py to avoid sending a flood of alerts.

The alerts seem to be sent only on the last try right before the script ACK the pubsub message even though it didn't successfully sync.

Please remove the additional complexity of saving state to the disk about if errors messages have been sent.

script/sync_prod_to_gcp/submissions_to_gcp.py
ntai-arxiv: This is to flag the error state marked.  

I'm using the /tmp/sync-to-gcp/ as a place to mark the notification sent so the following attempt doesn't send notification. The state file is cleared when the message is acked. 

It is a memory disk operation so it should be pretty quick. If I really wanted this to be atomic, I could use a lock file but it would be overkill when the event is already serialized by the pubsub.

PR #879, opened 2025-07-21
[HTML25-74] Guard missing ToC in HTML Papers

dginev:
A base for further ToC improvements, more gracefully accepts legitimate cases where the ToC does not get generated.

Example: older articles which are simply a list of paragraphs organized together.


norbusan (approved):
(good to have the "ignore whitespace" setting which makes checking the changes easy)

LGTM

PR #936, opened 2025-11-03
ARXIVCE-4082: Check whether elements exist before attempting to add listener. 

DavidLFielding:
Eliminate javascript error when opt-in is disabled.

Will test again on develop server once approved.

### DIFF
--- a/browse/templates/cookies.html
+++ b/browse/templates/cookies.html
@@ -120,7 +120,7 @@ preferences.
   // Generate UUID
   const regBtn = document.getElementById('opt-in-registration');
   if (regBtn) {
-    document.getElementById('opt-in-registration').addEventListener('click', event => {
+    regBtn.addEventListener('click', event => {
     event.preventDefault();
     const uuid = crypto.randomUUID(); // Generate a unique ID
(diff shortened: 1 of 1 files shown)


browse/templates/cookies.html
```diff
@@ -119,19 +118,24 @@ <h2>Opt-in Data Collection for Research (Cornell University)</h2>
...
-  });
+    const uuid = crypto.randomUUID(); // Generate a unique ID
+    document.cookie = `opt-in-tracking=${uuid}; Path=/; Max-Age=31536000`; // 1 year
+    checkOptInParticipation(); // Update UI
+    });
+  }
+
+  const optOutBtn = document.getElementById('opt-out');
+  if (optOutBtn) {
+      document.getElementById('opt-out').addEventListener('click', event => {
```
dginev:
```suggestion
      optOutBtn.addEventListener('click', event => {
```

dginev (changes requested):
a couple of DRY lines

browse/templates/cookies.html
DavidLFielding: I eliminated extra getElementById in two spots.

PR #967, opened 2026-04-01
Fixes /login to use AUTH_SERVER

bdc34:
This was fixed in arxiv-base.

Chanegs arxiv-base dependency to use `master` instead of a specific revision.

Sets uv to do a dependency cool down of 1 week with `exclude-newer`

Updates other dependencies.


dginev (approved):

ntai-arxiv (approved):
Excellent.

PR #972, opened 2026-04-06
Fix WCAG 2.2 accessibility audit findings

SBBCornell:
Addresses accessibility issues identified in the WCAG 2.2 audit that relate to this repo. All code changes done by Claude. Manually tested at https://arxiv-browse-accessibility-demo-874717964009.us-central1.run.app/ with accessibility checking tools and VoiceOver when needed.

Audit reports: [Public report](https://docs.google.com/spreadsheets/d/1_NvzUNhq_p8vnyI9bMcdQ6FLksW9W80q2RVhRYQ6wnw/edit?gid=0#gid=0), [Limited access review](https://wa-reviews.hosting.cornell.edu/project/80?tab=issues)

- Fix BibTeX export button: change span to button for keyboard accessibility. Maintain same visible style.
- Add real bookmark icon alt text.
- Add proper heading structure
- Fix archive form selects: add fieldsets, legends, and labels
- Add button elements with ARIA to category taxonomy accordions
- Labs tabs: add proper tablist/tab/tabpanel ARIA roles and states
- Fix toggle switch focus indicators: add visible focus outlines
- Fix link color contrast: add underlines to links
- Fix report issue modal: add focus trapping and aria-modal
- Fix stats chart: add role="img" and aria-label to SVG
- Mobile menu: fixed ARIA errors
- Exception: NO changes made to the HTML pages. 3 changes are suggested (dark mode toggle element, modal popup focus trapping, and back-to-reference focus indicator). I'll discuss with Deyan.

More context on WCAG 1.4.1 Use of Color: even though the default browser link color used in arXiv browse does not pass color contrast requirements, I did not underline all links as the audit seems to suggest. I am interpreting "inline links" in the WCAG criteria as meaning a link embedded within running text. We have those links too, for example within an abstract description, and they need to be underlined. But most links in arxiv-browse actually exist in contexts where everything is a link (homepage wall of links, author lists) or where visual context makes them obviously interactive (abs sidebar links).  In those cases, even arXiv's non-compliant color should technically be OK without an underline. We will do better in a redesign, but in the current context underlining every link on the abs page or on the homepage adds a lot of visual noise that our users would definitely notice.

### DIFF
--- a/browse/static/css/arXiv.css
+++ b/browse/static/css/arXiv.css
@@ -23,8 +23,14 @@ a:hover {
   text-decoration: underline;
 }
 
-/* Underline links within text content for accessibility (WCAG 1.4.1) */
-p a, li a, td a, th a, dd a, dt a, blockquote a, figcaption a, .abstract a {
+/* Underline links within prose content for accessibility (WCAG 1.4.1) */
+/* Only in contexts where links appear within running text, not navigation or link lists */
+.abstract a,
+blockquote:not(.abstract) a,
+figcaption a,
+.submission-history p a,
+#content .description a,
+.accordion-body p a {
   text-decoration: underline;
 }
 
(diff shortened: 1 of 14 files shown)


dginev (approved):
This PR is a definite improvement and should be merged. We can add additional polish in follow-ups as needed.

I made a qucik testing run on an /abs/ page with NVDA and tab navigation and discovered one interesting detail:
 - The live production site navigates into the content of the "Labs tabs" boxes, but never speaks out the tab header names.
 - This PR improves on that, as the first/active tab header is now navigated into and pronounced. But I did not manage to get NVDA to keyboard navigate into the adjacent tabs. Something to keep in mind.

PR #986, opened 2026-06-23
Cache sub-20MB downloads at Fastly by serving HTTP 200 instead of HTTP 206

dginev:
### Motivation

This is a follow-up to a discussion we had at the Tuesday dev meeting, Jun 23. Some analysis suggested that we may not be caching small PDF files due to how Fastly deals with HTTP 206 responses.

### Details

`default_resp_fn` answered every `Range` request with `206 Partial Content`. Fastly (apparently) will not cache a 206 without segmented caching, so each range request (in-browser PDF viewers, and the CDN itself) was forwarded to the Cloud Run origin -- the dominant driver of the carrier-peering "Network Data Transfer Out" egress cost.

Production logs showed ~1-2 MB 206 PDF responses are ~69% of /pdf origin egress, with a 3.19x average repeat factor (~68% eliminable by caching).

So we can only emit HTTP 206 for objects larger than  Fastly's whole-object cache limit of 20 MB. Smaller objects are returned in full (HTTP 200) with Accept-Ranges, so Fastly caches the object once and synthesizes later range responses from the edge. Larger objects keep the 206 path for Fastly segmented caching.

### Testing

We need to do some request tracking with Fastly post-deploy to check if the caching behavior is as intended.

### DIFF
--- a/browse/config.py
+++ b/browse/config.py
@@ -85,6 +85,20 @@ class Settings(arxiv_base.Settings):
     The cache-control: max-age to set for /pdf /src /e-print /html pages. Max
     for this in the RFC is one year.
     """
+
+    FASTLY_CACHE_MAX_OBJECT_SIZE: int = 20 * 1024 * 1024
+    """Largest object, in bytes, that Fastly will cache as a single whole object.
+
+    Files at or below this size are returned in full (HTTP 200) even when the
+    client/CDN sends a Range header, so Fastly caches the whole object once and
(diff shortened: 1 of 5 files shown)


dginev:
Superseded by #987 — same branch/commit, repushed to the upstream repo (arXiv/arxiv-browse) instead of my dginev fork so the `build-push` CI job's GCP workload-identity secret is available (GitHub withholds secrets from fork PRs). Identical content.

PR #990, opened 2026-06-25
[RI-396] self-heal an empty/inconsistent listing page

dginev:
### Motivation

On **2026-06-25**, `/list/hep-th/new` (and `/recent`) served an **empty page** in
production for ~10 hours. A read-replica lag spike during the announcement produced a
single empty render, and because browse ships listings with `Surrogate-Control: max-age=604800` (one week),
Fastly **froze** that empty page until a manual purge. The trigger is hard to eliminate; on the other hand, 
the amplifier (long-caching a bad render) is fully in browse's control, and this fixes it.

### Root cause

`/new` and `/recent` derive their **counts** from `arXiv_updates` (+ `arXiv_in_category`)
but render their **items** from a separate query that also requires `Metadata.is_current = 1`.
During lag the counts can be non-zero while the items come back empty.

### PR approach

`get_listing` guards the empty/inconsistent case before caching:

- **Inconsistency → `503`.** When the counts promise items this page should show but none
  rendered, return an uncacheable `503` (`Surrogate-Control: max-age=0`, `Retry-After`)
  instead of caching a wrong page. For `/new` the check is on the **new** count specifically,
  not the replacement-polluted total: a day of "only replacements" is a legitimate page, 
  while "new submissions promised yet absent" (their fresh metadata still lagging while older replacements render) is the failure.
  Surrogate keys are preserved so it stays purgeable.
- **Empty volatile page → short cache.** An empty daily-volatile announce page (new/recent/current month)
   without the contradiction is cached for only  `LISTING_EMPTY_MAX_AGE` (default 5 min) so a transient empty
   self-heals instead of freezing for a week. Immutable historical months keep their long cache.
- **`stale-if-error` on every `/list` response** (`LISTING_STALE_IF_ERROR`, default 1 day).
  This is makes the `503` somewhat safe for UX: on any origin 5xx, Fastly serves the last good listing rather than showing users the error, for a grace period.

### Config

| setting | default | purpose |
|---|---|---|
| `LISTING_EMPTY_MAX_AGE` | 300 s | short TTL for an empty volatile listing |
| `LISTING_STALE_IF_ERROR` | 86400 s | error-grace for serving the last good listing on origin 5xx |

### DIFF
--- a/browse/controllers/__init__.py
+++ b/browse/controllers/__init__.py
@@ -13,7 +13,10 @@ from arxiv.identifier import Identifier
 from werkzeug.datastructures import Headers
 
 
-Response = Tuple[Dict[str, Any], int|status, Dict[str, Any]|Headers]
+# The body is usually a template-context dict, but a controller may also return
+# a ready-made body string (e.g. the list_page 503 served on a listing
+# inconsistency, which must not go through normal template rendering).
+Response = Tuple[Dict[str, Any]|str, int|status, Dict[str, Any]|Headers]
 
 
 def check_supplied_identifier(id: Identifier, route: str) -> Optional[Response]:
--- a/browse/controllers/__init__.py
+++ b/browse/controllers/__init__.py
@@ -13,10 +13,7 @@ from arxiv.identifier import Identifier
 from werkzeug.datastructures import Headers
 
 
-# The body is usually a template-context dict, but a controller may also return
-# a ready-made body string (e.g. the list_page 503 served on a listing
-# inconsistency, which must not go through normal template rendering).
-Response = Tuple[Dict[str, Any]|str, int|status, Dict[str, Any]|Headers]
+Response = Tuple[Dict[str, Any], int|status, Dict[str, Any]|Headers]
 
 
 def check_supplied_identifier(id: Identifier, route: str) -> Optional[Response]:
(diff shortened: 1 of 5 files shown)


browse/controllers/list_page/__init__.py
```diff
@@ -278,11 +311,35 @@ def get_listing(subject_or_category: str,
...
+            # Counts promise items but none rendered (replica-lag fingerprint):
+            # 503 so Fastly serves the last good copy via stale-if-error; keep the
+            # surrogate keys so it stays purgeable.
+            if current_app.config.get("ARXIV_LOG_DATA_INCONSTANCY_ERRORS"):
+                logger.error(
+                    "Listing inconsistency for %s/%s: counts promise items but "
+                    "none rendered (skip=%d, count=%d, items=%d). Likely "
+                    "read-replica lag; returning 503 instead of a wrong page.",
+                    list_ctx_id, time_period, skipn, count, len(listings))
+            response_headers["Surrogate-Control"] = "max-age=0"  # never cache
```
kyokukou: we might want to very short term cache it just so we dont overwhelm the db with tons of people trying to fetch a listing that isnt ready yet

browse/controllers/__init__.py
```diff
@@ -13,7 +13,10 @@
 from werkzeug.datastructures import Headers
 
 
-Response = Tuple[Dict[str, Any], int|status, Dict[str, Any]|Headers]
+# The body is usually a template-context dict, but a controller may also return
+# a ready-made body string (e.g. the list_page 503 served on a listing
+# inconsistency, which must not go through normal template rendering).
+Response = Tuple[Dict[str, Any]|str, int|status, Dict[str, Any]|Headers]
```
kyokukou: i believe the response object is used all across browse, may want to be thoughtful about changing it

kyokukou (approved):

PR #991, opened 2026-06-25
add category agnostic tags to list

kyokukou:
lets you clear fastly cache of a lists for a specific month or year, rather than just category by category

### DIFF
--- a/tests/listings/test_list_page.py
+++ b/tests/listings/test_list_page.py
@@ -865,6 +865,7 @@ def test_surrogate_keys(client_with_db_listings):
     assert "list-ym" in head
     assert "announce"  not in head
     assert "list-2005-06-nlin.SI" in head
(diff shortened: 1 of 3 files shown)


tests/listings/test_list_page.py
```diff
@@ -865,13 +865,15 @@ def test_surrogate_keys(client_with_db_listings):
     assert "list-ym" in head
     assert "announce"  not in head
     assert "list-2005-06-nlin.SI" in head
+    assert "list-2005-06" in head
```
dginev: You may want to add spaces to ensure you are getting the exact keys (since they overlap with e.g. the test above `list-2005-06-nlin.SI"`
```
assert " list-2005-06 " in " "+head+" "
assert " list-2005 "    in " "+head+" "
```

dginev (approved):

PR #993, opened 2026-06-26
[ARXIVCE-4426] spinout header/footer ; browse

dginev:
### Description

Replaces browse's legacy header/footer with the spinout design. 

- Consumes the shared chrome same-origin via `url_for('base.static', …)` — no vendored copy; 
- browse also serves the `/static/base/<BASE_VERSION>/` route on the main domain.

- Reusable chrome partials, persistent (localStorage) banner dismissal, no-JS search fallback, AAA-passing colors, mobile menu, design-system tokens.

- Removes the header breadcrumbs site-wide.

- Runtime: browse's deployed arxiv-base `master` branch must carry the chrome (BASE_VERSION 1.0.1).

- updates `Makefile` to also be based on `uv`

### Open question

The spinout announcement banner is tentatively set to expire in 2 weeks (July 15). Is that the correct date?

### Preview

browse-top
browse-bottom

PR #994, opened 2026-06-29
[ARXIVCE-4426] Spinout header/footer feedback (browse)

dginev:
### Description

Follow-up to #993 . Addresses live-testing feedback on the new spinout header/footer.

### Changes

- **Footer width** for browse: span the page width to match the abstract page margins (full-width, ~10px sides) instead of the centred 1200px grid.
- **Footer position:** move the announcement/promo banners *inside*
  `.flex-wrap-footer` so the sticky-footer `100vh` accounts for the banner height;
   previously the footer was pushed ~one banner-height below the viewport (visible in /abs).
- **HTML papers** — remove the `BETA` corner flag.
- **Accessibility (WAVE)** — fix the HTML reader nav-label hover contrast: 
  the span-wrapped labels (Report Issue / Back to Abstract / Download PDF)
  hovered black (failing contrast on the ink bar); 
  they now use `--header-hover-color` (Open Blue), consistent with "Why HTML?" and the icons.

### Preview

image

Screenshot from 2026-06-29 14-48-06

### DIFF
--- a/browse/static/css/arXiv.css
+++ b/browse/static/css/arXiv.css
@@ -388,6 +388,11 @@ main {
    min-height: 100vh;
    flex-direction: column;
  }
+ /* Spinout footer: span the page width like the abs-page #content (very minimal side
+    margins) instead of the centred 1200px grid. Browse-local override of the shared
+    chrome — the descendant selector beats it regardless of stylesheet order. */
+ .flex-wrap-footer .ds-site-footer { padding-left: 10px; padding-right: 10px; }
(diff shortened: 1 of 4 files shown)


dginev:
I did a quick survey of our options for "Log in" / "Log out" , it will require some more careful work so won't attempt it on a short notice. For now I can offer the constant "Log in" button, matching the current arxiv.org live site.

kyokukou (approved):

PR #1003, opened 2026-08-07
Hide author-contact labels (Affiliation:/Email:) in the arXiv HTML theme

dginev:
## What

Hides the `.ltx_contact_name` labels — "Affiliation:" and "Email:" — that render before each author affiliation / email in the HTML paper header.

## Why

The latexml-oxide worker runs with `--nodefaultresources`, so LaTeXML's default `LaTeXML.css` (which italicised `.ltx_contact_name`, and only hid it under the `.ltx_authors_multiline` layout) is no longer emitted. ar5iv 0.9.0 doesn't re-declare the class, so the raw labels leak through on every affiliation/email line — e.g. 5× "Affiliation:" + 1× "Email:" on arXiv:2607.16395v1.

## Change

One rule in the theme's `ar5iv-patches` layer:

```css
.ltx_contact_name { display: none; }
```

This hides both the "Affiliation:" and "Email:" labels (the institution text and the mailto link stand on their own). To keep the "Email:" label, scope it to `.ltx_role_affiliation .ltx_contact_name` instead.

Witness: https://dev.arxiv.org/html/2607.16395v1

Keeping this open for testing over the coming week.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/browse/static/css/arxiv-html-papers-theme-20260807.css
+++ b/browse/static/css/arxiv-html-papers-theme-20260807.css
@@ -637,6 +637,11 @@ Empirically, the xl >1280px breakpoint seems to be the first where a sidebar TOC
     margin: 1rem;
   }
 
+  /* Hide the author-contact labels ("Affiliation:", "Email:"). */
+  .ltx_contact_name {
+    display: none;
+  }
+
   /* Restore a visible text selection. ar5iv 0.9.0 newly tints ::selection with
      var(--note-highlight-color) (a11y.css), which the arXiv theme sets near-white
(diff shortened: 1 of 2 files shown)


dginev:
Folded in a second, unrelated commit: `uv.lock` bumps arxiv-base `dd15adc`→`2d29bd7` so `latexml_html_path` uses archive-qualified (squashed) keys, matching what the conversion worker now writes. Without it, old-style HTML ids 404 (e.g. `astro-ph/0303073v1`); new-style ids unaffected. Happy to split it out if you'd rather review separately.

carly-jones (approved):

PR #1011, opened 2026-09-23
sync_prod_to_gcp: verify the /ftp object before retiring it to /orig

norbusan:
retire_bucket_objects() moved a replacement's previous version with a bucket-side copy of the /ftp object, assuming that object was still the previous version. When the publish sync is delayed, that assumption breaks. A replacement whose PDF does not build is retried until an admin compile fix, and that fix's restore.pl -> sync_paper.sh path has already synced the NEW current version to /ftp. The copy then puts the new version into orig/<id>v<N-1>, and arxiv.org/src/<id>v<N-1> serves it.

Seen on 2510.02751, 2510.26915, 2602.08779, 2605.18010, 2606.23834 and 2609.16946. For 2510.02751 the publish message was retried 219 times overnight, then logged "tar.gz is moved to orig/.../2510.02751v1.tar.gz" with the md5 of the fixed v2.

Now the /ftp object is copied only if its md5 matches CIT's copy of the retired version (/data/orig/...vN, already in the sync plan). Otherwise that CIT file is uploaded to /orig instead. Also reload from_blob before comparing, since bucket.blob() carries no md5_hash.


dginev:
Thanks Norbert, the diagnosis is spot on, and comparing against CIT's `/data/orig` copy is the right approach. A few things before merge:

**Blocking**
1. **A failed upload still acks the message.** The new `upload()` / `md5_base64()` calls aren't wrapped the way the copy path is. If they raise, `submission_callback` catches the exception, sees only the earlier `True` verdict from the `.abs` retire, and acks. I reproduced it: `/ftp/<id>.abs` had already been moved to orig, nothing was re-uploaded, and there was no retry. Please add `try/except` with `add_verdict(False, ...)`. Ideally, also record a `False` verdict in `submission_callback`'s `except` around `sync_to_gcp`.
2. **Move `from_blob.reload()` up, before the `to_blob.exists()` branch.** `exists()` fetches only `fields=name`, so `from_blob.md5_hash` is still `None` at the "identical" check on L988, and that branch can never fire. `bucket.get_blob()` would replace exists() + reload() with a single call.
3. **Already-damaged objects aren't healed.** For the six papers, `orig/...v<N-1>` already exists, so the code goes into the `to_blob.exists()` branch and never looks at CIT. We need a one-off re-upload from CIT plus a Fastly purge of `/src/<id>v<N-1>`. It's also worth a sweep in case there are more than six: orig objects created by `copy_blob` should carry `localpath=/data/ftp/...` in their metadata. Better still, compare `to_blob` against CIT in that branch so a re-sync fixes it.

**Non-blocking**
- If the CIT file is missing, the code quietly falls back to the old blind copy. Please at least log a warning.
- The test fake always has an md5, so it wouldn't catch #2. The CI workflow only runs `tests/`, so this test never runs there. The `logging_json` stub isn't needed since it's in `requirements.txt`.
- The check also applies to `.abs`. If CIT's orig `.abs` normally differs from the bucket's `/ftp` copy byte-for-byte, every replacement will log this warning. Worth watching after deploy.

---
_This review was derived with Claude Opus 5.5 (via Claude Code) and checked by me._

norbusan:
> A failed upload still acks the message. 

This is true, and was there before. The new code just adds more "options" to fail ;-) I will add code to work around this.

> Move from_blob.reload() up

I will merge the  #1012 into this one here, then this is cleared up

> Already-damaged objects aren't healed

Yes, on purpose

PR #1012, opened 2026-09-23
sync_prod_to_gcp: reload the /ftp blob before comparing with an existing /orig

norbusan:
When the retire destination already exists, retire_bucket_objects() compares to_blob.md5_hash with from_blob.md5_hash. Only to_blob was reloaded. from_blob comes from bucket.blob(), which carries no metadata, so its md5_hash was always None. The "identical" case therefore never matched: a leftover /ftp copy of an already-retired version was never removed, and every such case was logged as "Both exists but contents not the same!".


dginev:
Thanks Norbert, correct fix; this covers #2 from my #1011 comment. Three things:

1. **Coordinate with #1011.** The code change merges cleanly, but with both PRs merged this PR's tests fail (`ValueError: not enough values to unpack (expected 4, got 3)`) because #1011 changes the tuple to include `cit_path`. Suggest folding this into #1011, or merging this first and rebasing #1011. Keep a single `from_blob` reload (or `get_blob()`) right after the `exists()` check, and one test module using this PR's fake, which models the missing md5 correctly.
2. **Let's run these in CI; nothing does today.** `python-app.yml` only runs `pytest ... tests`. The fake-based tests work unchanged with the locked browse env (pubsub, storage and logging-json are all in `uv.lock`, so the `logging_json` stub can go). I checked with `uv sync --locked` + pytest from the repo root. A separate step is enough:
   ```yaml
       - name: Test sync_prod_to_gcp unit tests
         run: uv run pytest script/sync_prod_to_gcp/test/test_retire_*.py
   ```
   Please target these files and not the whole directory: `test_submissions_to_gcp.py` / `test_sync_prod_to_gcp.py` need a real bucket and credentials. With this step, CI would have caught (1).
3. **"Identical" ≠ correct.** In the state the #1011 bug leaves behind (orig/v1 and /ftp both hold v2), this branch now deletes the correct /ftp object and reports "Both Exists" success. The corrupted orig stays, and the logs make it look healthy; before this PR they at least said "not the same". Once `cit_path` is available from #1011, please compare against CIT here too.

Nit: `projection='full'` isn't needed for md5/size; the default `noAcl` projection includes both.

---
_This review was derived with Claude Opus 5.5 (via Claude Code) and checked by me._

norbusan:
Close this and merge it into #1011 as suggested, this should fix the CI tests and some other inconsistencies.

